PR #119, opened 2023-08-03
Beam geometry

chenkasirer:
- Restored the `geometry` accessor property to simplify Beam API
- Geometry is now created on first call to `beam.geometry` rather than automatically. This allows creating a `TimberAssembly` in contexts where geometry is not available or not of interest (e.g. cadwork).

### Checklist


src/compas_timber/ghpython/components/CT_ShowAssembly/code.py
```diff
@@ -12,6 +12,6 @@ def RunScript(self, Assembly):
 
         Brep = []
         for beam in Assembly.beams:
-            Brep.append(Artist(beam.get_geometry(True)).draw())
+            Brep.append(Artist(beam.geometry).draw())
```
funkchaser, line 15: check also other GH components, there are some more `beam.get_geometry` calls

funkchaser (changes requested):
Looks otherwise good to me, looks simpler/neater, but haven't run any tests/demos with the new workflow. 
Can you also check if it requires updating the documentation/tutorial?

src/compas_timber/ghpython/components/CT_ShowAssembly/code.py
chenkasirer: Nice catch :) Fixed in Bake component. Seen anywhere else?

chenkasirer:
> n a

Updated `Beam`'s docstring. Couldn't find any references in the docs. However, `get_geometry` still shows as method of `Beam` (inherited from `Part`) I would have to sort that out in COMPAS core at some point.

BTW, since `compas_timber` continues to rely on unreleased changes to COMPAS, I added the exact commit I'm currently based on to `requriements.txt` and in order to test you'd need to install using `pip install -r requirements-dev.txt`. Maybe better in a new env so to not mess up any other working projects that use COMPAS LTS.

chenkasirer:
@funkchaser ping :)

src/compas_timber/parts/beam.py
```diff
@@ -435,6 +416,10 @@ def from_endpoints(cls, point_start, point_end, width, height, z_vector=None, ge
         line = Line(point_start, point_end)
         return cls.from_centerline(line, width, height, z_vector, geometry_type)
 
+    @classmethod
+    def from_brep(cls, brep):
+        pass
```
gonzalocasas, line 421: maybe `raise NotImplementedError` instead of passing?

requirements-dev.txt
```diff
@@ -1,4 +1,3 @@
-black
```
gonzalocasas, line 1: no `black`?

src/compas_timber/connections/t_butt.py
```diff
@@ -82,10 +82,10 @@ def cutting_plane(self):
         cfr = Frame(cfr.point, cfr.yaxis, cfr.xaxis)  # flip normal towards the inside of main beam
         return cfr
 
-    def restore_beams_from_keys(self, assemly):
+    def restore_beams_from_keys(self, assembly):  # TODO: fix typo
```
gonzalocasas, line 85: the typo seems fixed, I guess the comment can go?

gonzalocasas (approved):
LGTM!

src/compas_timber/parts/beam.py
chenkasirer: removed altogether. will be in it's own PR.

requirements-dev.txt
chenkasirer: was there twice! :0

PR #121, opened 2023-08-30
Compas main

chenkasirer:
Updated the `Data` interface in `Beam` and `Assembly` to comply with the changes in COMPAS core intended to be released in version 2.0.

### Checklist


src/compas_timber/assembly/assembly.py
obucklin, line 1: Yes serialize that asssembly data!

obucklin (approved):
Thanks Chen! I loked over the changes and mostly get what you are doing. You are right it is a good educaitonal opportunity. 

Maybe let me know if or when I should update my COMPAS or CT to use the new  serialization

requirements-dev.txt
```diff
@@ -16,7 +16,7 @@ pydocstyle
 pytest >=3.2
 pytest-mock
 sphinx_compas_theme >=0.13
-sphinx >=3.4
+sphinx ==7.1.2
```
gonzalocasas, line 19: is this still required? higher version of sphinx break the theme?

gonzalocasas (approved):
lgtm! 

this got me thinking we should probably already bump the version in COMPAS, so that other dependencies such as this one can declare it properly, because now it's somewhat inconsistent, this version only works with `compas@main` but it's nowhere visible

requirements-dev.txt
chenkasirer: yes `7.2.4` breaks the template, I opened https://github.com/compas-dev/sphinx_compas_theme/issues/20. 
Also Sphinx are seemingly dropping python versions like flies. `7.2.0` already doesn't support python 3.8 anymore..

gonzalocasas:
> we can't add it to `requirements.txt` before there's an actual release of COMPAS 2.0. Is that what you meant?

yes, i don't know what's the best option for doing this, cut a 2.0-alpha release of COMPAS and use that?

chenkasirer:
> > we can't add it to `requirements.txt` before there's an actual release of COMPAS 2.0. Is that what you meant?
> 
> yes, i don't know what's the best option for doing this, cut a 2.0-alpha release of COMPAS and use that?
> 
> 

I'd be all in for that. I feel like we shouldn't be too far from 2.0 anyways. @tomvanmele what do you say?

PR #137, opened 2023-10-09
Rework

chenkasirer:
This is proposal for a re-work I think would make compas_timber a bit less rigid/fragile when used in different contexts.
Major changes are listed below and address the tight coupling of parametric definition and geometry operations.

* Remove visualization geometry related attributes and methods from Beam.
* Make `Beam` somewhat immutable (at least its parametric definition)
* New approach to Features 
    * `CutFeature`, `DrillFeature` and `MillFeature` are higher level and refer to operations (gets closer to machining semantic)
    * previous idea `ParametricFeature` is now its own thing, separate from above features.
* `Beam.blank` replaces `Beam.shape` and returns a parametric blank (box) which includes extensions needed for joinery, if any.
* Geometry creation for visualization purposes is externalized to a `BrepGeometryConsumer` which "consumes" a `TimberAssembly` and returns a (generator) collection of the Breps for visualization.

Further consumers could be `MachiningConsumer`, `CadworkConsumer` and others. 

API for visualization looks something like this:
```python3
assembly = TimberAssembly()
assembly.add_beam(b1)
assembly.add_beam(b2)

LMiterJoint.create(assembly, b1, b2)

viewer = App()

for result in BrepGeometryConsumer(assembly).result:
    # result has pairs of `Beam` and `Brep` objects i.e `result.beam`
    viewer.add(result.geometry)
viewer.run()
```
image

Changes in GH workflow:
* visualization of `Beam` blanks already in `BeamFromCurve` component

beam

* visualization of assembly with features already in the `Assembly` component
assembly


jonashaldemann (approved):
Approved as discussed, I totally agree with the concept

PR #140, opened 2023-11-14
Planning

chenkasirer:
> Note: this is based on `rework` relevant changes are in [this commit](https://github.com/gramaziokohler/compas_timber/commit/7e996770f153518a2b027d2fb2f98563c140304f)

Moved over `BuildingPlan` from `monosashi` (with some modifications) and implemented a simple sequence generator.

Make an assembly:
```python
assembly = TimberAssembly()
b1 = Beam(Frame.worldXY(), length=2.0, width=0.1, height=0.1)
b2 = Beam(Frame.worldXY(), length=5.0, width=0.1, height=0.1)
assembly.add_beam(b1)
assembly.add_beam(b2)
```

Creates a sequence based on the same order of the beams in the assembly:
```python
from compas_timber.planning import SimpleSequenceGenerator

generator = SimpleSequenceGenerator(assembly)
plan = generator.result

print(f"plan has {len(plan)} steps")

for index, step in enumerate(plan):
    beam_key = step.element_ids[0]
    print(f"beam #{index}: {assembly.find_by_key(beam_key)}")

```

```bash
> plan has 2 steps
> beam #0: Beam 0.100 x 0.100 x 2.000 at Frame(point=Point(0.0, 0.0, z=0.0), xaxis=Vector(x=1.0, y=0.0, z=0.0), yaxis=Vector(x=0.0, y=1.0, z=0.0))
> beam #1: Beam 0.100 x 0.100 x 5.000 at Frame(point=Point(0.0, 0.0, z=0.0), xaxis=Vector(x=1.0, y=0.0, z=0.0), yaxis=Vector(x=0.0, y=1.0, z=0.0))
```


mhelmrei:
Suggest: Add `grouping method` for multiple parts to be contained in one Step.
Suggest: Add Actor _'CNC Maschine'_
Suggest: Add Instruction Type _'Image'_

src/compas_timber/planning/sequencer.py
```diff
@@ -0,0 +1,274 @@
...
+        List of cad element ids which are associated with the step.
+
+    """
+
+    def __init__(
+        self,
+        element_ids,
+        location=None,
+        obj_filepath=None,
+        step_id=None,
```
chenkasirer, line 145: not really needed. add in a derived class if needed.

src/compas_timber/planning/sequencer.py
```diff
@@ -0,0 +1,274 @@
...
+    ):
+        super(Step, self).__init__()
+        self.element_ids = element_ids or []
+        self.location = location or Frame.worldXY()
+        self.obj_filepath = obj_filepath
+        self.step_id = step_id
+        self.instructions = instructions or []
+        self.is_built = is_built
+        self._actor = actor
+
```
chenkasirer, line 158: add `objects_held` - list of beam keys which are being held by the robot during the execution of this step.
add `priority` - steps with the same priority value can be interchangeable.

src/compas_timber/planning/sequencer.py
```diff
@@ -0,0 +1,274 @@
...
+        List of instructions which make up the step.
+    element_ids : list(int)
+        List of cad element ids which are associated with the step.
+
+    """
+
+    def __init__(
+        self,
+        element_ids,
+        location=None,
```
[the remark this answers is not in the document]
chenkasirer: maybe keep it. If all we do is visualize using OBJs, we can completely disregard the assembly itself.
All we need for visualization is a frame and an obj.

tests/compas_timber/test_sequencer.py
jckenny59, line 1: looks good to me 👍

jckenny59 (approved):
Looks good to me! 👍

PR #141, opened 2023-11-15
frame.data instead of live frame in Beam.data

chenkasirer:
* Bumped COMPAS dependency to `2.0.0a2`.
* Matched serialization of `Beam` to `Part` so that both could be interchangeably used when deserializing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Installing dependencies
         run: |
-          curl -o compas.tar.gz -LJO https://pypi.debian.net/compas/COMPAS-2.0.0a1.tar.gz
+          curl -o compas.tar.gz -LJO https://pypi.debian.net/compas/COMPAS-2.0.0a2.tar.gz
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
       - name: Installing dependencies
         run: |
-          curl -o compas.tar.gz -LJO https://pypi.debian.net/compas/latest
+          curl -o compas.tar.gz -LJO https://pypi.debian.net/compas/COMPAS-2.0.0a1.tar.gz
```
jckenny59, line 49: I am just curious, why does it map dependencies to Compas-2.0.0a1 and not the latest?

chenkasirer: Recently a `2.0.0.a2` has been released where some Brep related stuff has been moved around, which breaks different things in `compas_timber`.
We already actually already had `2.0.0.a1` pinned in the [requirements.txt](https://github.com/gramaziokohler/compas_timber/blob/main/requirements.txt), but ironpython workflow uses this different mechanism to get it's dependencies and there it would just grab the latest.

gonzalocasas (approved):
lgtm!

PR #213, opened 2024-02-07
Lbutt extend cross

chenkasirer:
* Merged the functionality of `DecomposeBeam` and `ShowBeamFrame` and removed the latter.
* Added visualization for beams faces and their normals to `DecomposeBeam`.
* Added two flags that can be used when assigning L-butt joints in a multi-beam joint situation:
    * `modify_cross` when set to `False` (default `True`) tells CT no to modify the cross beam (which might already be joined)
    * `reject_i` when set to `True` (default `False`) `BeamJoiningError` is raised when the main and cross meet in an I topology. I is generally allowed with L-butt but might not be desired when the cross is already joined at the same end.
   

When both are left as default, the L-butt behavior is as before.

image

image

### Checklist


src/compas_timber/connections/l_butt.py
```diff
@@ -79,14 +84,18 @@ def joint_type(self):
         return "L-Butt"
 
     def get_main_cutting_plane(self):
-        angles_faces = beam_side_incidence(self.main_beam, self.cross_beam)
-        cfr = min(angles_faces, key=lambda x: x[0])[1]
+        assert self.main_beam and self.cross_beam
+
+        index, cfr = self.get_face_most_ortho_to_beam(self.main_beam, self.cross_beam, ignore_ends=False)
+        if index in [5, 6]:  # end faces
```
obucklin, line 90: my concern is, what happens if someone wants to make a butt joint with beam incidence angle of < 45deg? In this case, the joint will have one of the ends as `cfr` or `index` will be 5 or 6, even though the user wants to join in the old way with just the 4 main faces. 

Should we add this as a kwarg in the GH component? it could be a cutoff angle with default of 45.

chenkasirer: a sketch here would help. 

but if I get it right though, what you're suggesting sounds like an l-miter, no?
otherwise, when considering only the 4 main faces you would get something like this:
image

jonashaldemann (approved):

PR #244, opened 2024-03-20
Feature priorities

chenkasirer:
* Features have now priorities (workaround for nasty issue applying certain brep subtractions before trimming)
* `Beam.attributes` now gets serialized.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


obucklin (commented):
Should `FeaturePriority` be an int instead so we can set the order in case of multiple overlapping features?

gonzalocasas:
I think it is already one, even if there's a class to wrap it, it's only acting as an int enum.

chenkasirer:
@gonzalocasas @obucklin I'm gonna close this one for now as this was more of a workaround and I'm not so happy about this solution. If I see there's no other way, I'll open it again.

PR #245, opened 2024-03-22
these fixes are needed  for the hangar model to work

obucklin:
These allow the hangar model to work. the one in butt_joint was a variable I forgot to change.

the changes in the L-butt joint should not pose any problem. the functions that I moved to outside the `try` should not fail.

- [`x`] Bug fix in a **backwards-compatible** manner.

### Checklist

### DIFF
--- a/src/compas_timber/connections/l_butt.py
+++ b/src/compas_timber/connections/l_butt.py
@@ -121,10 +121,9 @@ class LButtJoint(ButtJoint):
         if self.features:
             self.main_beam.remove_features(self.features)
         start_main, start_cross = None, None
-
+        main_cutting_plane = self.get_main_cutting_plane()[0]
+        cross_cutting_plane = self.get_cross_cutting_plane()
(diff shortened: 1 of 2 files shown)


src/compas_timber/connections/l_butt.py
```diff
@@ -121,10 +121,9 @@ def add_features(self):
         if self.features:
             self.main_beam.remove_features(self.features)
         start_main, start_cross = None, None
-
+        main_cutting_plane = self.get_main_cutting_plane()[0]
+        cross_cutting_plane = self.get_cross_cutting_plane()
```
chenkasirer, line 125: why were these moved out of the `try` block?

chenkasirer:
@obucklin what about this one? all in all it looked fine to me. Is that something we want to still merge? I was just wondering about moving these two lines out of the `try..except` this is normally necessary for the debug information feature to work.

PR #242, opened 2024-03-19
Beam_faces_match_btlx

obucklin:
Changed `Beam.faces` to match position and orientation of BTLx `ReferenceSide`. 

BTLx `ReferenceSide` indexing begins with 1, so the ReferenceSide index will be the `Beam.faces` index +1.

Beam.faces[i] = ReferenceSide: i+1

I also changed the `BTLx.frame` to match the BTLx `Part_Ref` base Frame.

### What type of change is this?

- [ ] Bug fix in a **backwards-compatible** manner.
- [ ] New feature in a **backwards-compatible** manner.
- [x] Breaking change: bug fix or new feature that involve incompatible API changes.
- [ ] Other (e.g. doc update, configuration, etc)

### Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [x] I ran all tests on my computer and it's all green (i.e. `invoke test`).
- [x] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [ ] I added new functions/classes and made them available on a second-level import, e.g. `compas_timber.datastructures.Beam`.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added necessary documentation (if appropriate)

PR #247, opened 2024-04-24
beam frame consistent with btlx reference frame

chenkasirer:
WIP

image
image

### Checklist

### DIFF
--- a/src/compas_timber/elements/beam.py
+++ b/src/compas_timber/elements/beam.py
@@ -1,5 +1,6 @@
 import math
 
+import compas
 from compas.geometry import Box
 from compas.geometry import Brep
 from compas.geometry import Frame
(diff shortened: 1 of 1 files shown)


obucklin (commented):
Hey is the goal to change the `beam.frame` to coincide with the BTLx `PartRef`. I would be a bit concerned that will screw up a bunch of stuff. Does it make sense to have the `beam.blank_frame` take that role?

obucklin:
Have you looked at this recently? Given that at least FEA and maybe other funtionalities rely on centerlines, I propose we keep the `beam.frame` on the nonextended centerline, and this as the `beam.blank_frame`. I had an earlier branch (PR?) that did this, and I tested it a fair amount. I would try to make sure that the `beam.faces` and `BTLx.reference_planes` at least refer to the same surfaces, even if the `frame.point` is different. keeping a clear distinction between beam and blank aka beam box and BTLx may be complicated and we should make sure the difference is clear

chenkasirer:
@obucklin alright, let's give it a try as you suggested.
I opened https://github.com/gramaziokohler/compas_timber/pull/265
which will replace this PR and is very similar to as https://github.com/gramaziokohler/compas_timber/pull/242, hope it's alright I reimplemented.

I'll do it in small chunks so keep it bite-sized.

PR #258, opened 2024-06-24
Recalculate Trigger for Model GH component

obucklin:
This adds the ability to trigger CreateGeometry on the CT_Model GH component. The component will mostly function as before, except that when the CreateGeometry input parameter is set to False, it maintains the last geometry created, instead of reverting to `beam.blanks`. This allows the user to use a button input to trigger a recalculation of the geometry, which is useful when dealing with large models where each parameter change would otherwise trigger a new solution. 

We could consider adding a reset button if it turns out users want to look at the `beam.blanks` too.

### Checklist


requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 attrs >=17.4
 black >=22.12.0
 bump-my-version
-compas_invocations2
+compas_invocations
```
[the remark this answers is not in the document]
obucklin: ok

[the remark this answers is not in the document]
obucklin: switched it back

src/compas_timber/ghpython/components/CT_Model/code.py
```diff
@@ -112,71 +115,73 @@ def get_joints_from_rules(self, beams, rules, topologies):
...
         scene = Scene()
-        for beam in Model.beams:
-            if CreateGeometry:
+        for beam in self.model.beams:
+            try:
                 scene.add(beam.geometry)
                 if beam.debug_info:
                     debug_info.add_feature_error(beam.debug_info)
-            else:
+            except Warning as w:
```
chenkasirer, line 179: where is a `Warning` thrown?

src/compas_timber/ghpython/components/CT_Model/code.py
```diff
@@ -112,71 +115,73 @@ def get_joints_from_rules(self, beams, rules, topologies):
...
-                beam_pair_ids = set([id(beam) for beam in beams_to_pair])
-                if beam_pair_ids in handled_beams:
-                    continue
-                try:
-                    joint.joint_type.create(Model, *beams_to_pair, **joint.kwargs)
-                except BeamJoinningError as bje:
-                    debug_info.add_joint_error(bje)
-                else:
-                    handled_beams.append(beam_pair_ids)
+        if create_geometry:
```
chenkasirer, line 129: maybe reverse this logic (i.e. `if not create_geometry: return`)? the indentation is already quite deep plus it makes it hard to see the change in this PR..

src/compas_timber/ghpython/components/CT_Model/code.py
```diff
@@ -112,71 +115,73 @@ def get_joints_from_rules(self, beams, rules, topologies):
                         )
         return joints
 
-    def RunScript(self, Beams, JointRules, Features, MaxDistance, CreateGeometry):
-        if not Beams:
+    def RunScript(self, beams, joint_rules, features, max_distance, create_geometry):
```
chenkasirer, line 118: nice :)

src/compas_timber/ghpython/components/CT_Model/code.py
```diff
@@ -24,6 +24,9 @@
 
 
 class ModelComponent(component):
+    def __init__(self):
+        self.model = TimberModel()
```
chenkasirer, line 28: wouldn't this just create an empty `TimberModel` every time the component is run? shouldn't the beams and joint etc. be added anyways (even if geometry is not calculated)

obucklin: I added this so that every the model would persist, even if the inputs changed. I don't know of another way to do this.

src/compas_timber/ghpython/components/CT_Model/code.py
obucklin: I never really used `try except`. I changed this to `FeatureApplicationError`, but I'm not sure if that is correct either.

src/compas_timber/ghpython/components/CT_Model/code.py
obucklin: nah this should only run if `create_geometry` is true, but there is more that needs to run after this.

src/compas_timber/ghpython/components/CT_Model/code.py
obucklin: Ok the other way would be to have the scene be persistent... you think that would be better?

src/compas_timber/ghpython/components/CT_Model/code.py
obucklin: Ok the `beam.geometry` returns the `beam.blank`anyways, so that was redundant anyways. deleted the `try/except`

PR #262, opened 2024-06-28
add plate element

obucklin:
This adds the plate element to Compas_timber. This is concieved as an extrusion with a boundary curve(Polyline) and a thickness. A vector input serves to switch the side towards which the plate is extruded. This has been tested with the Drill feature to ensure the Brep is oriented correctly and the boolean subtraction functions properly. Certain attributes are not yet implemented fully, such as the bounding box. TimberModel has been updated to include the `TimberModel.plates` property. 

I will continue to change other modules to remove references to `beam` in exchange for more generic `element` types when appropriate. 

### Checklist

invokes don't seem to be working on mac, I will run these when I get back to my PC.


src/compas_timber/elements/plate.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # ==========================================================================
+    # Alternative constructors
+    # ==========================================================================
+
+    @staticmethod
+    def _create_shape(outline, vector):
+        brep = Brep.from_extrusion(outline, vector)
+
+        return brep
```
chenkasirer, line 207:
```suggestion
        return Brep.from_extrusion(outline, vector)
```

src/compas_timber/elements/plate.py
```diff
@@ -0,0 +1,242 @@
...
+        -------
+        :class:`compas.datastructures.Mesh`
+            The collision geometry of the element.
+
+        """
+        return self.compute_aabb.to_mesh()
+
+    # ==========================================================================
+    # Alternative constructors
+    # ==========================================================================
```
chenkasirer, line 201: move down or remove this section header as there are currently no alternative constructors

src/compas_timber/model/model.py
```diff
@@ -45,18 +48,26 @@ def __from_data__(cls, data):
...
         super(TimberModel, self).__init__()
         self._beams = []
+        self._plates = []
         self._walls = []
         self._joints = []
         self._topologies = []  # added to avoid calculating multiple times
 
     def __str__(self):
-        return "TimberModel ({}) with {} beam(s) and {} joint(s).".format(self.guid, len(self.beams), len(self.joints))
+        return "TimberModel ({}) with {} beam(s), {} plate(s) and {} joint(s).".format(
```
chenkasirer, line 57: not necessarily related to this PR, but this is getting long. maybe we can format this similarly to `Brep`?
https://github.com/compas-dev/compas/blob/6e85276ce10d485e6b15f895b127c72a6862c7b4/src/compas/geometry/brep/brep.py#L153

src/compas_timber/ghpython/components/CT_Plate/code.py
```diff
@@ -0,0 +1,87 @@
...
+                    update_rhobj_attributes_name(guid, "category", c)
+
+                plates.append(plate)
+                scene.add(plate.shape)
+
+        geo = scene.draw()
+
+        return plates, geo
+
+    def _get_guid_and_geometry(self, line):
```
chenkasirer, line 77: could be in a future PR, but maybe we add this to the `ghpython_helpers` module

chenkasirer (changes requested):
Generally looks good, added some comments on the code.
Two general question/requests:
- please add unittests for the new `Plate`
- have you had a look at https://github.com/BlockResearchGroup/compas_model/blob/main/src/compas_model/elements/plate.py? maybe it's not right for us but might be worth it to have a look.

obucklin:
> * have you had a look at https://github.com/BlockResearchGroup/compas_model/blob/main/src/compas_model/elements/plate.py? maybe it's not right for us but might be worth it to have a look.

I had not seen that. the two polygons as faces seems like a good way to do it. I will try it out and see how it works. My only concern is whether their shape (a mesh) works with the CT features. Maybe you know...

src/compas_timber/elements/plate.py
obucklin: done

src/compas_timber/elements/plate.py
```diff
@@ -0,0 +1,242 @@
...
+        super(Plate, self).__init__(**kwargs)
+        if not outline.is_closed:
+            raise ValueError("The outline points are not coplanar.")
+        self.outline = outline
+        self.thickness = thickness
+        self.features = []
+        self.attributes = {}
+        self.attributes.update(kwargs)
+        self.debug_info = []
+        self.frame = Frame.from_points(outline.points[0], outline.points[1], outline.points[-2])
```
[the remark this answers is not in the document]
obucklin: this takes the first point as frame.point and then the two adjacent points to make the frame.

src/compas_timber/ghpython/components/CT_Plate/code.py
obucklin: ok I will leave it here for now

src/compas_timber/ghpython/components/CT_Model/code.py
```diff
@@ -112,27 +112,37 @@ def get_joints_from_rules(self, beams, rules, topologies):
...
+        for element in Elements:
+            if element is None:
+                continue
+            element.remove_features()
+            element.debug_info = []
+            Model.add_element(element)
+            if element.__class__.__name__ == "Plate":
+                plates.append(element)
+            elif element.__class__.__name__ == "Beam":
+                beams.append(element)
```
chenkasirer, line 138: I'd rather do:
`if isinstance(element, Plate)`

or better would be:
if element.is_plate

or even better than that would be:
we override `Model.add_element` and check it there

chenkasirer:
```python3
    def add_element(self, element, **kwargs):
        node = super(TimberModel, self).add_element(element, **kwargs)

        if isinstance(element, Beam):
            self._beams.append(element)
        elif isinstance(element, Wall):
            self._walls.append(element)
        elif isinstance(element, Plate):
            self._plates.append(element)
        else:
            raise NotImplementedError("Element type not supported: {}".format(type(element)))
        return node
```

src/compas_timber/model/model.py
```diff
@@ -45,18 +48,26 @@ def __from_data__(cls, data):
...
 
     @property
     def beams(self):
         # type: () -> list[Beam]
         return self._beams
 
+    @property
+    def plates(self):
+        # type: () -> list[Plate]
+        return self._plates
```
chenkasirer, line 69: `self._plates` should be declared in `TimberModel.__init__()`

src/compas_timber/ghpython/components/CT_Model/code.py
```diff
@@ -162,18 +172,18 @@ def RunScript(self, Beams, JointRules, Features, MaxDistance, CreateGeometry):
...
             if CreateGeometry:
-                scene.add(beam.geometry)
-                if beam.debug_info:
-                    debug_info.add_feature_error(beam.debug_info)
+                scene.add(element.geometry)
+                if element.debug_info:
+                    debug_info.add_feature_error(element.debug_info)
             else:
-                scene.add(beam.blank)
+                scene.add(element.blank)
```
chenkasirer, line 186: `if hasattr(element, "blank"): ...`

chenkasirer (approved):
LGTM!

PR #267, opened 2024-07-17
Jack cut feature

chenkasirer:
* Example implementation of a merged `JackRafterCut` and the `CutFeature`. Including client code in `L-Miter`.

I'm making an effort to make this change as gradual as possible. Meaning the two systems live side-by-side. For that purpose I created the new `JackRafterCut` in a new package `_fabrication` as otherwise there are some circular import issues.

image

### Checklist

### DIFF
--- a/src/compas_timber/_fabrication/jack_cut.py
+++ b/src/compas_timber/_fabrication/jack_cut.py
@@ -12,8 +12,7 @@ from compas.geometry import is_point_behind_plane
 from compas.geometry import Vector
 from compas.geometry import Rotation
 
-from compas.tolerance import TOL
-
+from compas_timber.elements import FeatureApplicationError
 
 if not compas.IPY:
     from typing import TYPE_CHECKING
@@ -206,10 +205,11 @@ class JackRafterCut(BTLxProcess):
         # type: (Brep, Beam) -> Brep
         cutting_plane = self.plane_from_params_and_beam(beam)
         try:
-            return geometry.trimmed(cutting_plane, TOL.absolute)
+            return geometry.trimmed(cutting_plane)
         except BrepTrimmingError:
-            raise Exception(
-                # TODO: use FeatureApplicationError
+            raise FeatureApplicationError(
+                cutting_plane,
+                geometry,
                 "The cutting plane does not intersect with beam geometry.",
             )
 
--- a/src/compas_timber/_fabrication/jack_cut.py
+++ b/src/compas_timber/_fabrication/jack_cut.py
@@ -1,24 +1,17 @@
 import math
 
-import compas
-from compas.geometry import Plane
+from compas.geometry import BrepTrimmingError
 from compas.geometry import Frame
 from compas.geometry import Line
-from compas.geometry import BrepTrimmingError
-from compas.geometry import intersection_line_plane
-from compas.geometry import distance_point_point
-from compas.geometry import angle_vectors_signed
-from compas.geometry import is_point_behind_plane
-from compas.geometry import Vector
+from compas.geometry import Plane
 from compas.geometry import Rotation
+from compas.geometry import Vector
+from compas.geometry import angle_vectors_signed
+from compas.geometry import distance_point_point
+from compas.geometry import intersection_line_plane
+from compas.geometry import is_point_behind_plane
 
 from compas_timber.elements import FeatureApplicationError
-
-if not compas.IPY:
-    from typing import TYPE_CHECKING
-
-    if TYPE_CHECKING:
-        from compas_timber.elements import Beam
 
 
 class BTLxProcess(object):
--- a/src/compas_timber/_fabrication/jack_cut.py
+++ b/src/compas_timber/_fabrication/jack_cut.py
@@ -13,33 +13,8 @@ from compas.geometry import is_point_behind_plane
 
 from compas_timber.elements import FeatureApplicationError
 
-
-class BTLxProcess(object):
-    """Base class for BTLx processes.
-
-    Attributes
-    ----------
-    ref_side_index : int
-        The reference side, zero-based, index of the beam to be cut. 0-5 correspond to RS1-RS6.
-    """
-
-    def __init__(self, ref_side_index):
-        self.ref_side_index = ref_side_index
-
-
-class OrientationType(object):
-    """Enum for the orientation of the cut.
-
-    Attributes
-    ----------
-    START : int
-        The start of the beam is cut away.
-    END : int
-        The end of the beam is cut away.
-    """
-
-    START = 0
-    END = 1
+from .btlx_process import BTLxProcess
+from .btlx_process import OrientationType
 
 
 class JackRafterCut(BTLxProcess):
--- a/src/compas_timber/_fabrication/jack_cut.py
+++ b/src/compas_timber/_fabrication/jack_cut.py
@@ -10,10 +10,12 @@ from compas.geometry import angle_vectors_signed
 from compas.geometry import distance_point_point
 from compas.geometry import intersection_line_plane
 from compas.geometry import is_point_behind_plane
+from compas.tolerance import TOL
 
 from compas_timber.elements import FeatureApplicationError
 
 from .btlx_process import BTLxProcess
+from .btlx_process import BTLxProcessParams
 from .btlx_process import OrientationType
 
 
@@ -36,6 +38,8 @@ class JackRafterCut(BTLxProcess):
         The vertical angle of the cut. 0.1 < inclination < 179.9.
 
     """
+
+    PROCESS_NAME = "JackRafterCut"  # type: ignore
 
     @property
     def __data__(self):
@@ -67,6 +71,10 @@ class JackRafterCut(BTLxProcess):
     ########################################################################
     # Properties
     ########################################################################
+
+    @property
+    def params_dict(self):
+        return JackRafterCutParams(self).as_dict()
 
     @property
     def orientation(self):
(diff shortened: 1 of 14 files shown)


src/compas_timber/elements/beam.py
```diff
@@ -467,7 +485,28 @@ def remove_blank_extension(self, joint_key=None):
...
+        # type: (int) -> compas.geometry.PlanarSurface
+        """Returns the requested side of the beam as a parametric planar surface.
+
+        Parameters
+        ----------
+        side_index : int
+            The index of the reference side to be returned. 0 to 5.
+
+        """
+        # TODO: maybe this should be the default representation of the ref sides?
```
papachap, line 498: I believe so as well. Using a parametric planar surface for the ref_side makes it more comprehensible i think.

src/compas_timber/elements/beam.py
```diff
@@ -406,6 +420,7 @@ def _create_shape(frame, xsize, ysize, zsize):
 
     @reset_computed
```
papachap, line 421: This resets the cached attributes of the class?

src/compas_timber/_fabrication/jack_cut.py
```diff
@@ -0,0 +1,276 @@
...
+        self.start_y = start_y
+        self.start_depth = start_depth
+        self.angle = angle
+        self.inclination = inclination
+
+    @property
+    def orientation(self):
+        return self._orientation
+
+    @orientation.setter
```
papachap, line 75: what is the benefit of using setters and private properties for this class?

src/compas_timber/_fabrication/jack_cut.py
```diff
@@ -0,0 +1,276 @@
...
+        The start of the beam is cut away.
+    END : int
+        The end of the beam is cut away.
+    """
+
+    START = 0
+    END = 1
+
+
+class JackRafterCut(BTLxProcess):
```
papachap, line 20: Where does the serialization of the btlx process parameters happen? shouldn't there be a class method for that? Also does this mean that header attributes are being removed?

src/compas_timber/connections/l_miter.py
papachap, line 1: With these changes does the l_mitter factory file gets removed?

papachap (approved):

papachap:
In case there are two btlx processes that are correlated (i.e StepJoint, StepJointNotch) and share most most of the parameters should the process be implemented as:

- two separate classes in the same module (_fabrication/step_joint) and generate the parameters twice if they both need to be applied in the case of a joint with two beams (one on main_beam and one cross_beam).

or


- as one class containing all the combined parameters with a boolean for applying the step or the step_notch

_(I attached the parameters for each one of the processes for the example of the StepJoint/StepJointNotch)_
Screenshot 2024-07-23 102032
Screenshot 2024-07-23 102229

src/compas_timber/_fabrication/jack_cut.py
```diff
@@ -0,0 +1,274 @@
...
+        orientation = cls._calculate_orientation(ref_side, plane)
+
+        point_start_x = intersection_line_plane(ref_edge, plane)
+        if point_start_x is None:
+            raise ValueError("Plane does not intersect with beam.")
+
+        start_x = distance_point_point(ref_edge.point, point_start_x)
+        angle = cls._calculate_angle(ref_side, plane, orientation)
+        inclination = cls._calculate_inclination(ref_side, plane, orientation)
+        return cls(orientation, start_x, start_y, start_depth, angle, inclination, ref_side_index=ref_side_index)
```
papachap, line 169: should the ref_side_index be in the this class constructor, since it is not a parameter of the class?

src/compas_timber/elements/beam.py
```diff
@@ -252,7 +265,7 @@ def __str__(self):
     # ==========================================================================
 
     def compute_geometry(self, include_features=True):
```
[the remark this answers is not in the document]
chenkasirer: exactly. `Beam` now creates its own geometry as a result as its `blank` + features.

src/compas_timber/elements/beam.py
chenkasirer: yes. some `Element` has some internal instance attributes which are cached after the first time they are created. For example `Element._geometry` is one. These need to be cleared when changes are made which might result in new geometry. This decorator takes care of that.

src/compas_timber/_fabrication/jack_cut.py
chenkasirer: this an object oriented thing that's strongly related to the encapsulation principle. In statically typed languages where private access is enforced you'd typically always need setters and getters. In python, since nothing is enforced by the language, it can get blurry.

There's different reasons why you'd want to use properties in python. In this particular case, I used properties and their setters for input validation. Since BTLx has a clear specification for the allowed value ranges for each parameter it seemed like a good place to validate those.

src/compas_timber/_fabrication/jack_cut.py
chenkasirer: You're right, that's missing. I introduced this in one of the last commits, might have not pushed yet.

I'm still very much debating though. I'd like to extract the XML schema specific stuff out of these classes, but they are nevertheless related. In the meantime, I created a `JackRafterCutParams` class whose sole responsibility is to populate the parameter values from `JackRafterCut` in they appropriate key-value pair according to the specification.

chenkasirer:
> In case there are two btlx processes that are correlated (i.e StepJoint, StepJointNotch) and share most of the parameters, should the process be implemented as:
> 
> * two separate classes in the same module (_fabrication/step_joint) and generate the parameters twice if they both need to be applied in the case of a joint with two beams (one on main_beam and one cross_beam).
> 
> or
> 
> * as one class containing all the combined parameters with a boolean for applying the step or the step_notch
> 
> _(I attached the parameters for each one of the processes for the example of the StepJoint/StepJointNotch the common parameters should have the same values if a joint between two beams is created)_ Screenshot 2024-07-23 102032 Screenshot 2024-07-23 102229

here I'd implement both separately. could be in the same module, but definitely as two separate classes. If the parameters of one of them is a subset of the other's then you could use inheritance to avoid having duplication code. 

Additionally, if it makes sense, we could make a factory method which creates matching pairs of `StepJoint` and `StepJointNotch` from the necessary collection of parameters.

src/compas_timber/_fabrication/jack_cut.py
chenkasirer: `ref_side_index` is an attribute of the parent class, as all processes need this parameter. notice `JackRafterCut` knows nothing about it, it gets it as a `kwarg` and passes it through to the parent in its call to `super().__init__()`.

src/compas_timber/connections/l_miter.py
chenkasirer: yup. that's the intention. `LMiterJoint` becomes the de-facto factory as it directly assigns the processes/features to the respective beams.

src/compas_timber/_fabrication/jack_cut.py
chenkasirer: it's in the latest commit https://github.com/gramaziokohler/compas_timber/pull/267/commits/aec27952f88eb1eb127eb97cfde185a936d68ada

src/compas_timber/_fabrication/jack_cut.py
```diff
@@ -0,0 +1,276 @@
...
+        start_depth = 0.0
+        ref_side = beam.ref_sides[ref_side_index]  # TODO: is this arbitrary?
+        ref_edge = Line.from_point_and_vector(ref_side.point, ref_side.xaxis)
+        orientation = cls._calculate_orientation(ref_side, plane)
+
+        point_start_x = intersection_line_plane(ref_edge, plane)
+        if point_start_x is None:
+            raise ValueError("Plane does not intersect with beam.")
+
+        start_x = distance_point_point(ref_edge.point, point_start_x)
```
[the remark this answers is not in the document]
chenkasirer: why? :)

src/compas_timber/elements/beam.py
chenkasirer: alright then I'd keep a mental note of maybe removing the frame-like ref sides. and we get rid of them in the future

chenkasirer:
@papachap I think I addressed all of your comments. 
I think it is currently not necessary to partially replace the processes just for the L-Miter. I'd leave it like this for now and make the swap from joint factories to process/features once we have all processes properly implemented.

If it's all the same to you, I'd go ahead an merge this one. Just hit approve if it's all good, otherwise feel free to leave some more comments.

src/compas_timber/_fabrication/jack_cut.py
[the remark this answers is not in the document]
papachap: ref_edge is a compas Line, no? in this case shouldn't it be ref_edge.start? I thought it was a typo or smth ;p

papachap (approved):
@chenkasirer Perfect, thanks! This was helpful! I'll try to do a pull request before next week for the `StepJoint` and `StepJointNotch` based on your implementation of the `JackRafterCut` and your feedback. We can discuss on that once I have it :)

papachap (approved):

src/compas_timber/elements/beam.py
obucklin: this comment was clarified in another PR, you can ignore my comment

obucklin (commented):
generally looks good. I think we just need to test the generation of the two angles with various plane angles.

PR #290, opened 2024-10-01
Gh bake rework

jonashaldemann:
Merged both Baking Components into one that bakes Beams & Plates. The Rhino Layers on which the elements should be baked can now also be defined.

image

### Checklist


src/compas_timber/ghpython/components/CT_Bake_BoxMap/code.py
```diff
@@ -14,44 +16,47 @@
...
         if not model:
             self.AddRuntimeMessage(Warning, "Input parameters Model failed to collect any Beam objects.")
             return
 
-        if not bake:
+        if not bake:  # Exit early if bake is False
             return
 
+        # Save current Grasshopper document context
+        ghdoc = sc.doc
```
chenkasirer, line 32: not sure if you need to do that. If I'm not mistaken `ghdoc` is always available in component scope.

src/compas_timber/ghpython/components/CT_Bake_BoxMap/code.py
```diff
@@ -14,44 +16,47 @@
...
 
-                for brep, frame in zip(breps, frames):
-                    guid = ActiveDoc.Objects.Add(brep)
-                    boxmap = self.create_box_map(frame, dimx, dimy, dimz)
-                    ActiveDoc.Objects.ModifyTextureMapping(guid, 1, boxmap)
         finally:
             rs.EnableRedraw(True)
 
+            # Restore document context back to Grasshopper
+            sc.doc = ghdoc
```
chenkasirer, line 56: Although I'm sure `rs.EnableRedraw()` shouldn't cause any issues, I would do `sc.doc=ghdoc` before. Just to be safe.

src/compas_timber/ghpython/components/CT_Bake_BoxMap/code.py
```diff
@@ -78,7 +78,31 @@ def create_box_map(pln, sx, sy, sz):
         dx = Interval(-sx * 0.5, sx * 0.5)
         dy = Interval(-sy * 0.5, sy * 0.5)
         dz = Interval(-sz * 0.5, sz * 0.5)
+        if swap_uv:
```
chenkasirer, line 81: so is it ok that there's only one `swap_uv`? this will flip the coordinates for both beams and plates right? can it be user might want to flip for just one of them?

chenkasirer:
@jonashaldemann looks generally good, although I did leave a few comments.

Now I'm not strictly against this re-work but keep in mind that as the selection of different element types, the number of input fields of this component is going to grow. This might end up being not very nice. But not sure having multiple components is the best solution either. Just something to keep in mind.

jonashaldemann:
> @jonashaldemann looks generally good, although I did leave a few comments.
> 
> Now I'm not strictly against this re-work but keep in mind that as the selection of different element types grows, the number of input fields of this component is going to grow. This might end up being not very nice. But not sure having multiple components is the best solution either. Just something to keep in mind.

Ok @chenkasirer, I see your objection. Maybe we can discuss it next week so I can estimate how many types there might be. Possibly the way is to create separate "Bakeries" but add the Layer Name Input to help assigning the Parts to Rhino Layers.

PR #310, opened 2024-10-28
added generic fastener element

obucklin:
This adds a generic fastener element which sets the stage for carious connectors like brackets and steel plates. We may want to add another generic 'dowel' class or we could include screws, nails, dowels, etc here under fastener. Heirarchy TBD.

### Checklist

### DIFF
--- a/src/compas_timber/elements/fasteners/fastener.py
+++ b/src/compas_timber/elements/fasteners/fastener.py
@@ -1,0 +1,200 @@
+import math
+
+from compas.geometry import Box
+from compas.geometry import Brep
+from compas.geometry import Frame
+from compas.geometry import Line
+from compas.geometry import PlanarSurface
+from compas.geometry import Plane
+from compas.geometry import Point
+from compas.geometry import Vector
+from compas.geometry import add_vectors
+from compas.geometry import angle_vectors
+from compas.geometry import bounding_box
+from compas.geometry import cross_vectors
+from compas_timber.utils import intersection_line_plane
+from compas.tolerance import TOL
+from compas_model.elements import reset_computed
+
+from .features import FeatureApplicationError
+from .timber import TimberElement
+
+
+class Fastener(TimberElement):
+    """
+    A class to represent timber fasteners (screws, dowels, brackets).
+
+    Parameters
+    ----------
+    elements : list(:class:`~compas_timber.parts.Element`)
+        The elements that are connected with this fastener.
+
+    Attributes
+    ----------
+    frame : :class:`~compas.geometry.Frame`
+        The coordinate system (frame) of this fastener.
+    elements : list(:class:`~compas_timber.parts.Element`)
+        The elements that are connected with this fastener.
+
+    """
+
+    OPPOSING_SIDE_MAP = {
+        0: 2,
+        2: 0,
+        1: 3,
+        3: 1,
+        4: 5,
+        5: 4,
+    }
+
+    @property
+    def __data__(self):
+        data = super(Fastener, self).__data__
+
+        return data
+
+    def __init__(self, elements, **kwargs):
+        super(Fastener, self).__init__(elements=elements, **kwargs)
+        self.elements = elements
+        self.features = []
+        self.attributes = {}
+        self.attributes.update(kwargs)
+        self.debug_info = []
+
+    def __repr__(self):
+        # type: () -> str
+        element_str = ["{} {}".format(element.type, element.key) for element in self.elements]
+        return "Fastener({})".format(", ".join(element_str))
+
+    # ==========================================================================
+    # Computed attributes
+    # ==========================================================================
+
+    @property
+    def is_fastener(self):
+        return True
+
+    @property
+    def shape(self):
+        # type: () -> Brep
+        assert self.frame
+        return self._create_shape(self.frame, self.beams)
+
+    @property
+    def key(self):
+        # type: () -> int | None
+        return self.graph_node
+
+    def __str__(self):
+        element_str = ["{} {}".format(element.type, element.key) for element in self.elements]
+        return "Fastener connecting {}".format(", ".join(element_str))
+
+    # ==========================================================================
+    # Implementations of abstract methods
+    # ==========================================================================
+
+    def compute_geometry(self):
+        # type: (bool) -> compas.geometry.Brep
+        """Compute the geometry of the fastener.
+
+        Returns
+        -------
+        :class:`compas.geometry.Brep`
+
+        """
+        raise NotImplementedError
+
+
+    def compute_aabb(self, inflate=0.0):
+        # type: (float) -> compas.geometry.Box
+        """Computes the Axis Aligned Bounding Box (AABB) of the element.
+
+        Parameters
+        ----------
+        inflate : float, optional
+            Offset of box to avoid floating point errors.
+
+        Returns
+        -------
+        :class:`~compas.geometry.Box`
+            The AABB of the element.
+
+        """
+        raise NotImplementedError
+
+    def compute_obb(self, inflate=0.0):
+        # type: (float | None) -> compas.geometry.Box
+        """Computes the Oriented Bounding Box (OBB) of the element.
+
+        Parameters
+        ----------
+        inflate : float
+            Offset of box to avoid floating point errors.
+
+        Returns
+        -------
+        :class:`compas.geometry.Box`
+            The OBB of the element.
+
+        """
+        raise NotImplementedError
+
+    def compute_collision_mesh(self):
+        # type: () -> compas.datastructures.Mesh
+        """Computes the collision geometry of the element.
+
+        Returns
+        -------
+        :class:`compas.datastructures.Mesh`
+            The collision geometry of the element.
+
+        """
+        return self.shape.to_mesh()
+
+    # ==========================================================================
+    # Alternative constructors
+    # ==========================================================================
+
+
+
+    @staticmethod
+    def _create_shape(frame, beams):
+        # type: (Frame,  list[TimberElement]) -> Brep
+        raise NotImplementedError
+
+    # ==========================================================================
+    # Featrues
+    # ==========================================================================
+
+    @reset_computed
+    def add_features(self, features):
+        # type: (Feature | list[Feature]) -> None
+        """Adds one or more features to the fastener.
+
+        Parameters
+        ----------
+        features : :class:`~compas_timber.parts.Feature` | list(:class:`~compas_timber.parts.Feature`)
+            The feature to be added.
+
+        """
+        if not isinstance(features, list):
+            features = [features]
+        self.features.extend(features)  # type: ignore
+
+    @reset_computed
+    def remove_features(self, features=None):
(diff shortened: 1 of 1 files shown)


obucklin:
i'm not sure what 

> "I added new functions/classes and made them available on a second-level import, e.g. compas_timber.datastructures.Beam" means, so I clicked it

obucklin:
Now with 100,000,000,000,000,000,000% more unit tests!

src/compas_timber/_fabrication/step_joint.py
```diff
@@ -75,7 +75,7 @@ def __init__(
         tenon=False,
         tenon_width=40.0,
         tenon_height=40.0,
-        **kwargs
+        **kwargs,
```
papachap, line 78: There is this issue, where the formatter adds a trailing comma after kwargs in __init__ on long lines, causing an 'unexpected token' error in IronPython. I guess to avoid that we can add `# fmt: off` as a comment directly above the __init__ method, when line breaking happens to prevent from reformatting each time the formatter runs."

src/compas_timber/_fabrication/step_joint_notch.py
```diff
@@ -91,7 +91,7 @@ def __init__(
         mortise=False,
         mortise_width=40.0,
         mortise_height=40.0,
-        **kwargs
+        **kwargs,
```
papachap, line 94: same here

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,182 @@
...
+    # ==========================================================================
+
+    @property
+    def is_fastener(self):
+        return True
+
+    @property
+    def shape(self):
+        # type: () -> Brep
+        assert self.frame
```
papachap, line 62: I guess you need to define a way for getting the frame for the fastner from the two (or more) interacting elements.

papachap (commented):
> Now with 100,000,000,000,000,000,000% more unit tests! 

haha I see that Chen presentation yesterday really had an impact on you!

papachap (commented):
> > Now with 100,000,000,000,000,000,000% more unit tests! 
> 
> haha I see that Chen's presentation yesterday really had an impact on you!

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+
+    OPPOSING_SIDE_MAP = {
+        0: 2,
+        2: 0,
+        1: 3,
+        3: 1,
+        4: 5,
+        5: 4,
+    }
```
chenkasirer, line 30: what's this doing here?

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,182 @@
...
+
+    @property
+    def __data__(self):
+        data = super(Fastener, self).__data__
+
+        return data
```
chenkasirer, line 36: if there's nothing to add here just use super AKA no need for `__data__` here

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,182 @@
+from compas_model.elements import Element
+from compas_model.elements import reset_computed
+
+
+class Fastener(Element):
```
chenkasirer, line 5: I'm confused about this class. it re-defines some stuff from `Element` without implementing them. feels like whatever inherits from it should just inherit directly from `Element`.

Also why does a `Fastener` get features? I get it can have holes etc, but do we really need to represent them? 
In my opinion, we should be careful with the granularity level of what we represent and opt for very simple representation as a start.

chenkasirer: Another thought, should we maybe represent certain fasteners as just joints? for me that would fit because they would inflict drills and cuts to the beams they joint, kinda like the joints we already have. no?

src/compas_timber/_fabrication/step_joint.py
obucklin: @chenkasirer what is the best soluiton here?

src/compas_timber/elements/fasteners/fastener.py
obucklin: My thought is, they are applied as a type of joint. the joint would apply the processing aka Features as previous joint do, but the joint would also apply the `Fastener(Element)` objects as well.

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,164 @@
...
+        # type: () -> str
+        element_str = ["{} {}".format(element.__class__.__name__, element.key) for element in self.elements]
+        return "Fastener({})".format(", ".join(element_str))
+
+    # ==========================================================================
+    # Computed attributes
+    # ==========================================================================
+
+    @property
+    def is_fastener(self):
```
obucklin, line 39: @chenkasirer can we have each element type register itself with the generic Element class so we don't need an explicit `is_fastener()` property in the Element class?

src/compas_timber/_fabrication/step_joint.py
chenkasirer: what panos said ;)

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,165 @@
...
+    ----------
+    elements : list(:class:`~compas_timber.parts.Element`)
+        The elements that are connected with this fastener.
+
+    """
+
+    def __init__(self, elements, **kwargs):
+        super(Fastener, self).__init__(elements, **kwargs)
+        self.elements = elements
+        self.features = []
```
chenkasirer, line 24: what exactly are features of a fastener? our features are currently btlx aligned so not sure that fits here.

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,165 @@
...
+    Attributes
+    ----------
+    elements : list(:class:`~compas_timber.parts.Element`)
+        The elements that are connected with this fastener.
+
+    """
+
+    def __init__(self, elements, **kwargs):
+        super(Fastener, self).__init__(elements, **kwargs)
+        self.elements = elements
```
chenkasirer, line 23: why are the elements saved here? are these the elements (beams) this fastener connects?
if so then it's up to the model to describe and maintain this relationship.

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,165 @@
...
+        self.features = []
+        self.attributes = {}
+        self.attributes.update(kwargs)
+        self.debug_info = []
```
chenkasirer, line 27: why is this needed here?

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,165 @@
...
+        inflate : float
+            Offset of box to avoid floating point errors.
+
+        Returns
+        -------
+        :class:`compas.geometry.Box`
+            The OBB of the element.
+
+        """
+        raise NotImplementedError
```
chenkasirer, line 104: shouldn't these be implemented? if not then there's no reason to re-define these, they are defined in `Element` already.

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,165 @@
...
+        -------
+        :class:`compas.datastructures.Mesh`
+            The collision geometry of the element.
+
+        """
+        return self.shape.to_mesh()
+
+    # ==========================================================================
+    # Alternative constructors
+    # ==========================================================================
```
chenkasirer, line 120: there are no alternative constructors here

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @staticmethod
+    def _create_shape(frame, beams):
+        # type: (Frame,  list[TimberElement]) -> Brep
+        raise NotImplementedError
```
chenkasirer, line 125: I assume this is internally used by `compute_geometry` which itself is not implemented so perhaps this should be here.

src/compas_timber/elements/fasteners/fastener.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # ==========================================================================
+    # Featrues
+    # ==========================================================================
+
+    @reset_computed
+    def add_features(self, features):
```
chenkasirer, line 132: see comment above about features

chenkasirer (changes requested):
I am still confused about `Fastener` it looks like a partial copy of `Element` and `Beam` but not clear how it's used. From our whiteboard session I'm missing the `FastenerJoint` part and how it interacts with `FastenerElement`. Should we have another session to discuss?

src/compas_timber/_fabrication/step_joint.py
obucklin: we should do this on another PR

chenkasirer: why? isn't this breaking IronPython?

papachap: Guys that line already exists in main! @obucklin make sure to pull from main. I added it in one of the PR's I merged(double_cut). The issue was only on these two modules.

obucklin:
> > > I am still confused about `Fastener` it looks like a partial copy of `Element` and `Beam` but not clear how it's used. From our whiteboard session I'm missing the `FastenerJoint` part and how it interacts with `FastenerElement`. Should we have another session to discuss?
> > 
> > 
> > I don't see the need for a separate `FastenerJoint` ATM. do you?
> 
> ah I think what's happening is that you're doing all that in that other PR (BallFastenerJoiningBallWithTentaclesJoint). so maybe that's why I'm confused.

tryna keep it grainular

src/compas_timber/_fabrication/step_joint.py
```diff
@@ -76,7 +76,7 @@ def __init__(
...
```
chenkasirer, line 79:
```suggestion
        **kwargs
```

src/compas_timber/_fabrication/step_joint_notch.py
```diff
@@ -92,7 +92,7 @@ def __init__(
...
```
chenkasirer, line 95:
```suggestion
        **kwargs
```

chenkasirer (approved):
LGTM

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added new method `has_group` to `TimberModel`.
 * Added new method `get_elements_in_group` to `TimberModel`.
 * Added attribute `is_group_element` to `TimberElement`.
+* Added new `fasteners.py` module with new `Fastener` element type.
+* Added unit tests for `fasteners.py` module.
```
chenkasirer, line 39: no need for that in the changelog
```suggestion
```

PR #313, opened 2024-10-31
New_solver_workflow

obucklin:
We were talking about this for a while, so I went ahead and did it. I needed it to change for the fastener class I added in another PR.

### Checklist


obucklin:
This is a pretty big change, so it may break some stuff. We should test more before merging, but also not wait too long...

obucklin (commented):
I've tested this a fair amount, and think it's ready for a review.

obucklin:
@chenkasirer @papachap hey would you guys take a look at this please?

src/compas_timber/connections/solver.py
```diff
@@ -133,7 +135,7 @@ def find_topology(self, beam_a, beam_b, tol=TOLERANCE, max_distance=None):
 
         """
 
-        tol = self.TOLERANCE  # TODO: change to a unit-sensitive value
+        tol = ConnectionSolver.TOLERANCE  # TODO: change to a unit-sensitive value
```
chenkasirer, line 138: as we discussed, I think this solver (and especially future solvers) could indeed have state.
For example, this solver can get `tol` and `max_distance` upon instantiation. these don't really change throughout the run of the solver and thus can be instance state.

src/compas_timber/connections/solver.py
```diff
@@ -1,5 +1,6 @@
 import itertools
 import math
+import stat
```
chenkasirer, line 3: imported but not used, could that be?

src/compas_timber/design/workflow.py
```diff
@@ -34,6 +37,87 @@ def comply(self, beams):
...
+                    comply, pair = rule.comply(pair)
+                    if comply:
+                        match_found = True
+                        joint_defs.append(JointDefinition(rule.joint_type, pair, **rule.kwargs))
+                        break
+            if not match_found:
+                print(
+                    "Beam pairs could not be resolved by the rules: ",
+                    "({}, {})".format(list(pair)[0].key, list(pair)[1].key),
+                )  # TODO: add something to catch unresolved pairs
```
chenkasirer, line 118: since this is inside the library now, prints and such (feedback to user should be handled by the calling code that's user facing, AKA in the GH component)

src/compas_timber/design/workflow.py
```diff
@@ -77,10 +161,19 @@ def __repr__(self):
...
         try:
             beam_cats = set([b.attributes["category"] for b in beams])
-            return beam_cats == set([self.category_a, self.category_b])
+            comply = False
+            beams = list(beams)
+            if beam_cats == set([self.category_a, self.category_b]):
+                solver = ConnectionSolver()
+                if (
+                    self.joint_type.SUPPORTED_TOPOLOGY
+                    == solver.find_topology(beams[0], beams[1], max_distance=max_distance)[0]
```
chenkasirer, line 173: here you still use `find_topology` as an instance method, even though you made it static. I personally think it should be instance method again, but at the very least it should be used consistently

src/compas_timber/elements/timber.py
```diff
@@ -28,3 +28,9 @@ def is_plate(self):
     @property
     def is_wall(self):
         return False
+
+    def reset(self):
```
chenkasirer, line 32: please add docstring

chenkasirer (changes requested):
looks generally good. left some comments
also please fix the build, it's currently failing

src/compas_timber/connections/solver.py
obucklin: ooh VScode keeps adding stupid imports >_<

src/compas_timber/connections/solver.py
obucklin: good point. I will revert to an instance method

obucklin: should all the methods be instance methods then?

src/compas_timber/design/workflow.py
obucklin: What about raise warning()?

src/compas_timber/design/workflow.py
obucklin: I will revert to instance method. should all the `ConnectionSolver`all be instance methods?

obucklin:
ready for re-re-view

src/compas_timber/design/workflow.py
```diff
@@ -34,6 +37,87 @@ def comply(self, beams):
...
+                        break
+
+            if not match_found:
+                for rule in JointRule.get_topology_rules(rules):  # see if pair is used in a topology rule
+                    comply, pair = rule.comply(pair)
+                    if comply:
+                        match_found = True
+                        joint_defs.append(JointDefinition(rule.joint_type, pair, **rule.kwargs))
+                        break
+            if not match_found:
```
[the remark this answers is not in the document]
chenkasirer: @obucklin ping

chenkasirer (approved):
LGTM!

PR #320, opened 2024-11-11
Walls as containers for groups of elements

chenkasirer:
Added some new functionality to `TimberModel` which allows the use of `Wall` elements as groups containing other elements.

This is the first step of the bigger multi wall feature.

### Checklist


src/compas_timber/model/model.py
```diff
@@ -123,6 +123,105 @@ def element_by_guid(self, guid):
...
+            A generator of elements in the group.
+
+        """
+        # type: (str, Optional[callable]) -> Generator[Element, None, None]
+        if not self.has_group(group_name):
+            raise ValueError("Group {} not found in model.".format(group_name))
+
+        filter_ = filter_ or (lambda _: True)
+
+        group = next((group for group in self._tree.groups if group.name == group_name))
```
papachap, line 221: here you use the `next` function in case there is more than one group with the same group name? If yes, can't a potential group name conflict create more issues elsewhere?

tests/compas_timber/test_wall.py
```diff
@@ -0,0 +1,91 @@
...
+
+    assert model.has_group("wall1")
+    assert model.has_group("wall2")
+    assert list(model.get_elements_in_group("wall1", filter_=lambda b: b.is_beam)) == [beam_a, beam_b]
+    assert list(model.get_elements_in_group("wall2", filter_=lambda b: b.is_beam)) == [beam_c, beam_d]
+
+
+def test_group_does_not_exist(model):
+    with pytest.raises(ValueError):
+        list(model.get_elements_in_group("non_existent_group"))
```
papachap, line 75: can you explain why you need to convert it to a list here?

papachap (approved):
LGTM 👍 
(I left a couple of questions to understand better how some functions work.)

src/compas_timber/model/model.py
chenkasirer: no, multiple groups with the same name are not allowed.
see https://github.com/gramaziokohler/compas_timber/pull/320/files#diff-2231cbd16043765cae13ca138761d868f5b2bc3e264227a237f2dab618ee29d8R172

I'm using `next` here because I expect only one result from the inner expression (which returns a generator).

This is instead of doing something really verbose like
```python3
result_group = None
for group in self._tree.groups:
    if group.name == group_name:
        result_group = group
        break
```
just more compact and clean.

I could also do something like:
```python3
group = [group for group in self._tree.groups if group.name == group_name][0]
```
but I find that ugly

tests/compas_timber/test_wall.py
chenkasirer: mmh that's probably not necessary here. co-pilot's fault ;)

PR #322, opened 2024-11-12
Lap BTLx Feature // T-Butt & L-Butt

papachap:
This PR introduces a new `BTLxLap` processing feature and Updates the `T-Butt` and `L-Butt` joints so that they follow the new system of joint definition (ie the use of the alt constructors of the `JackRafterCut` and `Lap` classes).

This means that the `ButtJoint` module can eventually be removed.

Also this PR fixes a bug in the calculations of _angle_ and _inclination_ of the `JackRafterCut`.

### Checklist


papachap:
I haven't removed the `ButtJoint `module, although it has been completely removed from the T-Butt and L-Butt joints. Let me know if you would like to do this in this PR or in a separate one.

src/compas_timber/_fabrication/lap.py
```diff
@@ -0,0 +1,510 @@
...
+        :class:`~compas.geometry.Brep`
+            The resulting geometry after processing
+
+        """
+        # type: (Brep, Beam) -> Brep
+        box = self.volume_from_params_and_beam(beam)
+        lap_volume = Brep.from_box(box)
+        try:
+            return geometry - lap_volume
+        except BrepTrimmingError:
```
chenkasirer, line 429: don't think that's the right error here. check what `RhinoBrep.from_boolean_difference()` raises when it fails to produce anything. if I remember right it's `IndexError` (as trying to index the first result).

src/compas_timber/connections/t_butt.py
```diff
@@ -20,25 +21,54 @@ class TButtJoint(ButtJoint):
...
-        super(TButtJoint, self).__init__(main_beam, cross_beam, mill_depth, birdsmouth, **kwargs)
-
-    def restore_beams_from_keys(self, model):
-        """After de-serialization, restores references to the main and cross beams saved in the model."""
-        self.main_beam = model.element_by_guid(self.main_beam_guid)
-        self.cross_beam = model.element_by_guid(self.cross_beam_guid)
+    @property
+    def __data__(self):
+        data = super(TButtJoint, self).__data__
+        data["main_beam"] = self.main_beam
```
chenkasirer, line 43: commented on the new pr also. beams guids were used here to prevent serializing the same beam multiple times.
instead, we beam is serialized just once, as part of the model, then on de-serialization the instances are references back here by `restore_beams_from_keys`.

src/compas_timber/connections/t_butt.py
```diff
@@ -53,36 +83,57 @@ def add_extensions(self):
...
+        self.features = [main_feature]
+
+        # apply the pocket on the cross beam
         if self.mill_depth:
-            self.cross_beam.add_features(MillVolume(self.subtraction_volume()))
-        self.main_beam.add_features(trim_feature)
-        self.features = [trim_feature]
+            cross_cutting_planes = (
+                self.main_beam.ref_sides[self.main_beam_ref_side_index],
+                self.main_beam.ref_sides[(self.main_beam_ref_side_index + 2) % 4],
```
chenkasirer, line 128: is there a way to make sense of these values to help someone reading this code?
is this an opposing face or maybe adjacent? can `Beam.OPPOSING_SIDE_MAP` maybe be used?

src/compas_timber/fabrication/__init__.py
```diff
@@ -17,7 +18,7 @@
     "BTLxLap",
     "BTLxFrenchRidgeLap",
     "LButtFactory",
-    "TButtFactory",
+    # "TButtFactory",
```
chenkasirer, line 21: is it not imported anywhere? 
perhaps start by removing the registration of this factory in the original btlx package.

tests/compas_timber/test_jack_rafter_cut.py
```diff
@@ -0,0 +1,121 @@
...
+    # cut the start of the beam
+    normal = Vector(x=-0.996194698092, y=-0.0, z=-0.0871557427477)
+    plane = Plane(Point(x=460.346635340, y=445.167151490, z=473.942755901), normal)
+    instance = JackRafterCut.from_plane_and_beam(Frame.from_plane(plane), beam)
+
+    # assert tol.is_close(instance.start_x, 214.922)
+    # assert tol.is_close(instance.start_y, 0.0)
+    # assert tol.is_close(instance.angle, 113.344)
+    # assert tol.is_close(instance.inclination, 95.443)
+    # assert tol.is_close(instance.ref_side_index, 0)
```
chenkasirer, line 83: remove?

src/compas_timber/_fabrication/lap.py
papachap: I think the correct one would be `BrepError` which "Represents a generic error in the Brep context"

papachap:
Adjustments based on the comments from @chenkasirer  done 👍

src/compas_timber/connections/t_butt.py
```diff
@@ -1,12 +1,13 @@
...
+from compas_timber._fabrication import Lap
+from compas_timber.connections.utilities import beam_ref_side_incidence
 
 from .joint import BeamJoinningError
+from .joint import Joint
 from .solver import JointTopology
 
 
-class TButtJoint(ButtJoint):
+class TButtJoint(Joint):
```
chenkasirer, line 10: so `ButtJoint` is  not used anymore, right? 
if so, perhaps remove it?

src/compas_timber/_fabrication/btlx_process.py
```diff
@@ -148,3 +148,44 @@ class LimitationTopType(object):
...
+    face_limited_back : bool
+        Limit the back face.
+
+    Properties
+    ----------
+    limits : dict
+        The limits dictionary with values as a boolean.
+    """
+
+    EXPECTED_KEYS = ["FaceLimitedStart", "FaceLimitedEnd", "FaceLimitedFront", "FaceLimitedBack"]
```
chenkasirer, line 175: what is this for? is it used anywhere?

src/compas_timber/_fabrication/lap.py
```diff
@@ -0,0 +1,511 @@
...
+        self._machining_limits = machining_limits
+
+
+    ########################################################################
+    # Alternative constructors
+    ########################################################################
+
+    @classmethod
+    def from_two_planes_and_beam(cls, planes, beam, depth, ref_side_index=0):
+        """Create a Lap instance from two planes and a beam. The planes should be parallel to each other and their normals should be facing in the opposite direction.
```
chenkasirer, line 292: a couple of questions here
these two planes are the side walls of the lap, right?
since there's quite a strict requirement, would it maybe be more robust if caller provides just one of the planes and the desired distance between them? 
the we could create the second plane ourselves, that's quite trivial I guess.

src/compas_timber/_fabrication/lap.py
chenkasirer: check out https://github.com/compas-dev/compas/blob/46b286311ccf0212f65fa202684c978a1d6bf560/src/compas/geometry/brep/brep.py#L672

currently if no results come back from the boolean operation there'll be an `IndexError` occurring due to our implementation always trying to index the first result.

I agree this should throw some kind of a `BrepError` instead, but that'll need to be changed in core.

src/compas_timber/connections/l_butt.py
```diff
@@ -55,34 +63,59 @@ def __init__(
...
                 beams=self.beams, joint=self, debug_info="Beams are in I topology and reject_i flag is True"
             )
-        return super(LButtJoint, self).get_main_cutting_plane()
+        return ref_side_index
+
+    @property
+    def main_beam_opposing_side_index(self):
+        return self.main_beam.opposing_side_index(self.main_beam_ref_side_index)
+
+    def update_beams(self):
```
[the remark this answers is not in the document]
papachap: Yes, you are right.

src/compas_timber/_fabrication/btlx_process.py
papachap: Yes i am using it in the setter of the machining limits of the Lap object where I check if the parameter provided is a dictionary and after that i check if both the keys and values are what they should.

src/compas_timber/connections/t_butt.py
obucklin: The `ButtJoint` class was used primarily for common geometric operations like generating the boolean subtraction solid. If I understand correctly, the geometric features are generated by the `btlx_process`es, so `ButtJoint` can presumably be phased out.

src/compas_timber/_fabrication/lap.py
```diff
@@ -0,0 +1,511 @@
...
+        # type: (Beam) -> Brep
+
+        ref_side = beam.side_as_surface(self.ref_side_index)
+        p_origin = ref_side.point_at(self.start_x, self.start_y)
+
+        box_frame = Frame(p_origin, ref_side.frame.xaxis, -ref_side.frame.yaxis)
+        rot_angle = 90-self.angle if self.orientation == OrientationType.END else self.angle-90
+        box_frame.rotate(math.radians(rot_angle), box_frame.normal, point=box_frame.point)
+
+        box = Box(xsize=self.length, ysize=self.width, zsize=self.depth, frame=box_frame)
```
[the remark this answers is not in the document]
chenkasirer: discussed with @papachap, I agree we should at least be able to do what was possible before AKA arbitrary laps as a result of lap joinery.

a potential solution would be a `from_beam_beam()` constructor that follows similar logic as the current lap joint. see implementation by @jonashaldemann: https://github.com/gramaziokohler/compas_timber/blob/295c3ffad6a55ebf69fd55a691948e994dfceff2/src/compas_timber/connections/lap_joint.py#L128

src/compas_timber/connections/l_butt.py
```diff
@@ -55,34 +63,59 @@ def __init__(
         self,
         main_beam=None,
         cross_beam=None,
-        mill_depth=0,
+        mill_depth=None,
         small_beam_butts=False,
-        modify_cross=True,
+        modify_cross=False,
```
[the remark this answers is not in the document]
papachap: by mistake, changed it back

src/compas_timber/_fabrication/lap.py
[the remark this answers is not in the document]
papachap: This is true that it is a simplification. For sure It is wise to implement an additional alternative constructor as @chenkasirer suggested in order to allow for non box-like laps by using the two beams to extract the params.

Also the geometrical representation can be done with a polyhedron volume rather than a box.

In regards to the joint definitions of TButtJoint & LButtJoint though, this simplification existed from before as the inclination was set to 90 degrees by default. See: https://github.com/gramaziokohler/compas_timber/blob/295c3ffad6a55ebf69fd55a691948e994dfceff2/src/compas_timber/connections/t_butt.py#L86
But it doesn't require the beams to be orthogonal, it will just create a lap as the one you can see in the picture below.

image


The `_create_negative_volumes()` was used only for the `XHalfLapJoint`. 

Let me know if we want to keep this simplification to any of those joint definitions.

papachap:
> I worry that we are implementing overly-simplified geometric implementations of the BTLx Processings. I would prefer that we make these more complete so that we don't have to go back later. This is particularly true for the `Lap` Processing, which will lose functionality with this. I know this is currently only for the Butt Joints, but it's worth considering if we should just implement fully immediately.

I agree with your point, and I think in this case it can be resolved by implementing an additional alternative constructor. But in my opinion in some of the joint definitions this simplification could be valuable. Let me know what you guys think @chenkasirer  @obucklin

papachap:
@obucklin After discussing with @chenkasirer, we agreed it would be wise to split the Lap Processing development into two PRs. The main reason is that this PR’s primary goal is to update the old implementations, at least when it comes to machining processes. Since lap joints currently only allow for geometric representation without the ability to extract machining parameters, we can accept that the alternative constructor introduced here is mainly intended for butt joints.

In a follow up PR, we can develop a new constructor that allows for the generation of a more advanced Lap processing, that would handle lap joints and would be able to extract the necessary machining parameters. (And also have the lap be represented as a volume rather than as a box).

chenkasirer (approved):
LGTM
there's a follow-up task https://github.com/gramaziokohler/compas_timber/issues/332 to complete the Lap visualization together with the Half-laps

PR #327, opened 2024-11-14
T butt plate

obucklin:
This PR demonstrates the "off-the-shelf" fastener workflow. It includes a new generic `PlateFastener` class that allows flat plate fasteners to be serialized. It is implemented in the T-Butt Joint and can later be added to other joint types.

### Checklist


obucklin:
@chenkasirer @papachap ready for review, Thanks!!

src/compas_timber/connections/t_butt.py
```diff
@@ -44,19 +44,43 @@ def __data__(self):
...
-    def __init__(self, main_beam=None, cross_beam=None, mill_depth=None, **kwargs):
+    def __init__(self, main_beam=None, cross_beam=None, mill_depth=None, fastener=None, **kwargs):
         super(TButtJoint, self).__init__(**kwargs)
         self.main_beam = main_beam
         self.cross_beam = cross_beam
         self.main_beam_guid = kwargs.get("main_beam_guid", None) or str(main_beam.guid)
         self.cross_beam_guid = kwargs.get("cross_beam_guid", None) or str(cross_beam.guid)
         self.mill_depth = mill_depth
         self.features = []
+        if isinstance(fastener, type):
```
chenkasirer, line 55: so if I understand right, there are two mutually exclusive scenarios here:
1. a fastener instance is passed to `__init__`
2. a fastener class is passed as `fastener` 

it is a bad practice to allow passing different types as the same argument as it's confusing to read and hard to test and debug.

something like this would be preferable and maybe would achieve the same goal:
```python
self.base_fastener = None
if fastener is None:
    self.base_fastener = PlateFastener.default_t()
```

Another issue is that there's another scenario, de-serialization, where the fastener instance is already part in the model but not available yet in `__init__`, therefore it's necessary to save its guid and restore it in `restore_elements_from_guids`.

src/compas_timber/elements/fastener.py
```diff
@@ -12,6 +12,27 @@
 from compas_timber.utils import intersection_line_box
 
 
+class FastenerApplicationError(Exception):
+    """Raised when a feature cannot be applied to an element geometry.
```
chenkasirer, line 17: please update this description

src/compas_timber/elements/fastener.py
```diff
@@ -185,32 +207,34 @@ def shape(self):
...
+                self._shape = geometries[0]
+                for geometry in geometries[1:]:
+                    self._shape += geometry
         return self._shape
 
     @property
     def geometry(self):
         """returns the geometry of the interface in the model (oriented on the timber element)"""
-        return self.shape.transformed(Transformation.from_frame(self.frame))
+        return self.shape.transformed(Transformation.from_frame(self.frame)) if self.shape else None
```
chenkasirer, line 225: `geometry` is already implemented in `Element` and it calls `compute_geometry` (which you implemented in the parent class above)

src/compas_timber/elements/fasteners/plate_fastener.py
```diff
@@ -0,0 +1,340 @@
...
+    Attributes
+    ----------
+    geometry : :class:`~compas.geometry.Geometry`
+        The geometry of the fastener.
+    frame : :class:`~compas.geometry.Frame`
+        The frame of the fastener.
+
+    """
+
+    def __init__(self, shape=None, frame=None, angle=math.pi / 2, topology=None, interfaces=[], outline = None, thickness = None, **kwargs):
```
chenkasirer, line 50: looking at the three ways to instantiate this class, make me feel like there room for revising how `PlateFastener` is defined, i.e. what are the basic attributes which constitute a `PlateFastener`.

specifically this `__init__`, the class method `from_outline_thickness_interfaces_cutouts` and the class method `default_T` all create an instance and while they have some overlaps they potentially create very different versions of instances of this class.

A good way to go about it is to define the core attributes that are essential to the creation of a `PlateFastener` and accept them in `__init__()`. it should always be possible to create a basic instance of it via the default init.

There might be some additional, optional, arguments that specify the fastener further, these should be accepted by `__init__` as well, as optional kwargs.

All of the above mentioned attributes should either already exist in parent classe(s) (e.g. `frame`) and therefore passed to `super` or they should be declared in the `__init__` body.

subsequently alternative constructors should provided some or all of these attributes (at least the core ones) to `__init__`, normally without creating new attributes.

src/compas_timber/elements/fasteners/plate_fastener.py
```diff
@@ -0,0 +1,340 @@
...
+        :class:`~compas_timber.elements.PlateFastener`
+
+        """
+        plate_fastener = cls(**kwargs)
+        plate_fastener.outline = outline
+        plate_fastener.angle = angle
+        plate_fastener.thickness = thickness
+        plate_fastener.frame = frame
+        plate_fastener.interfaces = interfaces
+        plate_fastener.cutouts = cutouts
```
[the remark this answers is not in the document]
obucklin: the bare minimum woudl be outline and thickness

src/compas_timber/connections/t_butt.py
```diff
@@ -44,19 +44,43 @@ def __data__(self):
...
+            fastener = fastener.default_T(main_beam.width)
+        self.base_fastener = fastener
+        self.fasteners = []
```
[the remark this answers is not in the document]
obucklin: 'base_fastener' is like a fastener definition and it exists at world_XY origin. the `fasteners` are the instances of the fastener placed in the model

src/compas_timber/connections/t_butt.py
obucklin: I wanted to cover the case where there is no fastener added to the joint. I see two options:

- add a second argument to the Joint __init__() called `add_fastener` which takes a boolean. Assuming that we can add fasteners to almost any joint, this seems unwieldy.
-  pass an instance of the `PlateFastener` and change its dimensions in the `Joint` def. I can do this, but it would essentially mean creating a completely new fastener.
- We could also consider a new `FastenerDefinition` class which becomes the `joint.base_fastener`

src/compas_timber/elements/fastener.py
obucklin: this is not an `Element`

tests/compas_timber/test_t_butt_joint.py
```diff
@@ -18,3 +20,21 @@ def test_create():
     assert isinstance(instance, TButtJoint)
     assert instance.main_beam == B1
     assert instance.cross_beam == B2
+
+
+def test_create_with_fastener():
+    B1 = Beam.from_endpoints(Point(0, 5, 0), Point(10, 5, 0), z_vector=Vector(0, 0, 1), width=1, height=2)
```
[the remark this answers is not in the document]
obucklin: I changed the way geometry is handled so I think it is now ok. I think it won't work to serialize geometry inputs from rhino, but I guess that is normal

obucklin:
@chenkasirer i moved the errors module to this PR because I was getting the same circular import errors.

src/compas_timber/elements/fasteners/plate_fastener.py
```diff
@@ -0,0 +1,319 @@
...
+    @property
+    def shapes(self):
+        for interface in self.interfaces:
+            for shape in interface.shapes:
+                yield shape
+
+    def set_default(self, joint):
+        width_a = joint.beams[0].width
+        width_b = joint.beams[1].width
+        if joint.SUPPORTED_TOPOLOGY == 3:  # JointTopology.TOPO_T
```
chenkasirer, line 131: notice `SUPPORTED_TOPOLOGY` is a list now.
also if possible, use `JointTopology.TOPO_` instead of the actual value

obucklin: yeah I was getting circular import errors when importing `joint.SUPPORTED_TOPOLOGY` so I went with the `int` value.

obucklin:
@chenkasirer ready!

chenkasirer (approved):
lgtm

PR #340, opened 2024-12-05
X/T/L HalfLap rework & Lap Volume viz adjustment

papachap:
image

This PR concerns the rework of the `HalfLap` joints (X,T,L) so that they use the `Lap` process instead. For the time being, they can allow half laps for beams that are aligned (meaning that their `z_vectors` are either perpendicular or parallel). In the future, the `Lap` process should support non-aligned configurations. However, this requires first defining a clear logic for extracting the `BTLx parameters` of the `Lap` process from the given geometry.

In this PR, the generation of the lap volume was adjusted to fully utilize all relevant `BTLx parameters`, ensuring more accurate visualization of the `Lap` process.

### Checklist


papachap:
There’s quite a bit of repeated code across the X, T, and L joint modules. This might be a good chance to decide what we can generalize and make accessible at a higher level to avoid duplication and keep things cleaner. @chenkasirer @obucklin

src/compas_timber/_fabrication/lap.py
```diff
@@ -470,29 +479,142 @@ def volume_from_params_and_beam(self, beam):
...
+            The frames of the cut.
+        beam : :class:`compas_timber.elements.Beam`
+            The beam that is cut by this instance.
+
+        Returns
+        -------
+        list of :class:`compas.geometry.Frame`
+            The updated frames of the cut.
+        """
+        TOL.absolute=1e-3
```
chenkasirer, line 597: best not to change global state if possible

src/compas_timber/connections/l_halflap.py
```diff
@@ -62,43 +110,100 @@ def add_extensions(self):
...
+        assert self.beam_a and self.beam_b
+        start_a, start_b = None, None
         try:
-            main_cutting_frame = self.get_main_cutting_frame()
-            cross_cutting_frame = self.get_cross_cutting_frame()
+            start_a, end_a = self.beam_a.extension_to_plane(self.cutting_plane_a.to_plane())
+            start_b, end_b = self.beam_b.extension_to_plane(self.cutting_plane_b.to_plane())
+        except AttributeError as ae:
+            # I want here just the plane that caused the error
+            geometries = [self.cutting_plane_b] if start_a is not None else [self.cutting_plane_a]
```
chenkasirer, line 120: normally I don't care much for performance, but it's worth paying attention when some properties are called multiple times where the value shouldn't have changed and the calculation can be rather signification (like in this case the `self.cutting_plane_*`)

src/compas_timber/connections/l_halflap.py
```diff
@@ -1,14 +1,16 @@
...
+from compas_timber.connections.utilities import beam_ref_side_incidence_with_vector
 
 from .joint import BeamJoinningError
-from .lap_joint import LapJoint
+from .joint import Joint
 from .solver import JointTopology
 
 
-class LHalfLapJoint(LapJoint):
+class LHalfLapJoint(Joint):
```
chenkasirer, line 13: so are we sure we want to get rid of the `LapJoint` class? discussed with @obucklin as well?

src/compas_timber/connections/l_halflap.py
```diff
@@ -31,25 +33,71 @@ class LHalfLapJoint(LapJoint):
...
+        ref_side_dict = beam_ref_side_incidence(self.beam_a, self.beam_b, ignore_ends=True)
+        ref_side_index = max(ref_side_dict, key=ref_side_dict.get)
+        return self.beam_b.side_as_surface(ref_side_index)
+
+    @property
+    def cutting_plane_b(self):
+        # the plane that cuts beam_b as a planar surface
+        ref_side_dict = beam_ref_side_incidence(self.beam_b, self.beam_a, ignore_ends=True)
+        ref_side_index = max(ref_side_dict, key=ref_side_dict.get)
+        return self.beam_a.side_as_surface(ref_side_index)
```
chenkasirer, line 100: these look very similar to the ones above, are they the same.
if we don't want these in a parent class, maybe they could be moved into a new higher level function(s) in our joinery toolkit modules where they return what's needed (index of surface) instead of the dictionary.

Then they can be called once in `add_features` this will be more readable, will have less duplication code and will potentially be more performant as some of these properties end up being called multiple times.

src/compas_timber/connections/l_halflap.py
```diff
@@ -62,43 +110,100 @@ def add_extensions(self):
...
+        """Checks if the elements are compatible for the creation of the joint.
 
-        self.features = [main_volume, cross_volume, f_main, f_cross]
+        Raises
+        ------
+        BeamJoinningError
+            If the elements are not compatible for the creation of the joint.
+        """
+        # check if the beams are aligned
+        for beam in self.elements:
```
chenkasirer, line 191: does this really have to be done once for each beam?

src/compas_timber/connections/x_halflap.py
```diff
@@ -1,45 +1,174 @@
-from compas_timber.elements import MillVolume
+from compas.tolerance import TOL
+
+from compas_timber._fabrication import Lap
+from compas_timber.connections.utilities import beam_ref_side_incidence
+from compas_timber.connections.utilities import beam_ref_side_incidence_with_vector
```
chenkasirer, line 5: these should be importable from the 2nd level i.e. `from compas_timber.connections import ...` or relative import like the ones below could also work since we're in the  same package anyways

src/compas_timber/connections/l_halflap.py
papachap: no there is no need. I should be calculating the dot product directly from the normal of the frames of the two beams.

src/compas_timber/connections/l_halflap.py
papachap: Yeah, I think we can drop the current LapJoint class since it’s no longer relevant or needed. To keep things cleaner and avoid redundancy, we should combine the XHalfLapJoint, LHalfLapJoint, and THalfLapJoint into one class—maybe call it HalfLapJoint or just LapJoint (same thing that we are doing for the TenonMortiseJoint). Since JointTopology is already a list, we can handle all the topologies (X, L, T) in the same class and avoid repeating the same logic in three places.

@obucklin @chenkasirer what do you guys think?

src/compas_timber/connections/l_halflap.py
papachap: Yeah, there’s definitely some redundancy here. The first pair of properties is for getting the ref_side for the `Lap` Processing of each beam, and the second pair is for calculating the planes needed for the alternative constructor of the `Lap` class. Even though each pair uses a different function to figure out the incidence, the two properties in each pair use the same function, so there’s some code repetition.

I’m not really a fan of generalizing this into something like `get_ref_side_index()` or `get_cutting_plane()` since each joint has its own way of calculating these values. It could make things less flexible and harder to adapt for different joint types.

What I’d suggest is converting these properties into lists and iterating through the beams to calculate the planes for each one. That would clean up the code, cut down on duplication, and make it easier to read.

what do you think?

obucklin: as per our conversation on Tuesday, we should not get rid of the generic parent classes and we should generate topology specific implementations for each type. Please let us know when you make those changes of if you have questions or comments.

papachap:
@chenkasirer  @obucklin  Based on the things we discussed, I have refactored the implementation for the lap joints and this PR is ready for re-review. 
In it's current status it **allows** the vizual representation of non co-planar lap joints and warns the user that `BTLxProcessings` were not generated from the process.

src/compas_timber/connections/utilities.py
```diff
@@ -164,20 +162,20 @@ def are_beams_coplanar(beam_a, beam_b, tol=TOL):
     bool
         True if the beams are coplanar, False otherwise.
     """
+    if len(beams) != 2:
```
chenkasirer, line 165: if this function only supports 2 beams, why not keep it explicit by keeping the previous signature?

src/compas_timber/fabrication/lap.py
```diff
@@ -385,29 +377,36 @@ def _calculate_orientation(ref_side, planes):
...
+        return max(x_distances) if orientation == OrientationType.END else min(x_distances)
 
     @staticmethod
     def _calculate_angle(ref_side, planes):
         # vector rotation direction of the plane's normal in the vertical direction
         angle_vector = Vector.cross(ref_side.normal, planes[0].normal)
         return abs(angle_vectors_signed(ref_side.xaxis, angle_vector, ref_side.normal, deg=True))
 
+    @staticmethod
+    def _calculate_inclination(ref_side, planes, is_pocket):
```
chenkasirer, line 403: why does this accept a list of planes if always only the first is used?

tests/compas_timber/test_lap.py
```diff
@@ -15,344 +16,263 @@
...
+def tol():
+    return Tolerance(unit="MM", absolute=1e-3, relative=1e-3)
 
+
+def test_lap_for_pocket_from_frame(tol):
     centerline = Line(
         Point(x=30396.1444398, y=-3257.66821289, z=73.5839565671),
         Point(x=34824.6096086, y=-3257.66821289, z=73.5839565671),
     )
+    cross_cection = [60, 120]
```
chenkasirer, line 28: `cross_section`?

chenkasirer (changes requested):
looks good. still a few small issues

src/compas_timber/connections/t_butt.py
```diff
@@ -155,13 +155,15 @@ def add_features(self):
         # apply the pocket on the cross beam
         if self.mill_depth:
             cross_cutting_plane = self.main_beam.ref_sides[self.main_beam_ref_side_index]
-            lap_width = self.main_beam.height if self.main_beam_ref_side_index % 2 == 0 else self.main_beam.width
+            lap_length = self.main_beam.height if self.main_beam_ref_side_index % 2 == 0 else self.main_beam.width
```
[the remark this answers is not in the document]
papachap: I made a suggestion for making it less cryptinc by inroducing a method for getting the perpendicular and parallel dimension of a beam to a ref_side.
If it's still not clear enough, I could revise it.

chenkasirer (approved):
LGTM

PR #344, opened 2024-12-10
BTLx Feaature Component

obucklin:
Added a GH component that outputs element geometry filtered by element type.

### Checklist


src/compas_timber/ghpython/components/CT_BTLx_Feature/code.py
```diff
@@ -0,0 +1,66 @@
...
+            kwargs["ref_side_index"] = ref_side
+            for arg, val in zip(self.arg_names(), args):
+                if val is not None:
+                    kwargs[arg] = val
+
+            process = self.processing_type(**kwargs)
+            face = beam.ref_sides[ref_side]
+
+            line_scene = Scene()
+            line_scene.add(Line.from_point_direction_length(face.point, face.xaxis, beam.length))
```
papachap, line 44: Why don't you return the frame instead?

src/compas_timber/ghpython/components/CT_BTLx_Feature/metadata.json
```diff
@@ -0,0 +1,36 @@
...
+        "iconDisplay": 0,
+        "inputParameters": [
+            {
+                "name": "Beam",
+                "description": "Beam to apply BTLx feature to.",
+                "typeHintID": "none",
+                "scriptParamAccess": 0
+            },
+            {
+                "name": "Reference Face",
```
papachap, line 19: Maybe call this 'reference_side' or 'ref_side' instead so that it's coherent to the BTLx naming convention.

papachap (commented):
I think we should include a visualization showing the result of the BTLx process applied to the beam. Since FeatureDefinition contains both the feature and the element, we could call the feature.apply() method, which takes the beam geometry as a Brep and the beam itself as arguments, and return the modified geometry.

papachap:
Potentially we could even consider exporting a BTLx file that includes just the beam and the applied BTLx process.

PR #348, opened 2024-12-18
Tenon/Mortise BTLx Process

papachap:
image

This PR introduces the implementation of the `Tenon` and `Mortise` BTLx processes. While tenons and mortises should ideally be treated as features that can be added to joints rather than standalone joints themselves, a `TTenonMortiseJoint` has been introduced as a temporary solution.

This joint implementation provides a practical way to integrate tenon and mortise features until a clearer logic is defined for handling joint features separately within the system.

### Checklist


src/compas_timber/connections/t_tenon_mortise.py
```diff
@@ -0,0 +1,246 @@
...
+from compas_timber._fabrication import Tenon
+from compas_timber._fabrication.btlx_process import TenonShapeType
+from compas_timber.connections.utilities import beam_ref_side_incidence
+
+from .joint import BeamJoinningError
+from .joint import Joint
+from .solver import JointTopology
+
+
+class TTenonMortiseJoint(Joint):
```
chenkasirer, line 11: does this really need to be specific to T?
couldn't we just have a TenonMortiseJoint which you could apply to either T or L? (in terms of this implementation, assuming this is possible regardless of what's in `SUPPORTED_TOPOLOGY`)

src/compas_timber/_fabrication/mortise.py
```diff
@@ -0,0 +1,551 @@
...
+        :class:`compas.geometry.Brep`
+            The mortise volume.
+
+        """
+        # type: (Beam) -> Brep
+        assert self.length is not None
+        assert self.width is not None
+        assert self.depth is not None
+
+        TOL.RELATIVE = 1e-3
```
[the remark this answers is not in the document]
papachap: I anyway don't need it anymore. Noted though..

src/compas_timber/connections/t_tenon_mortise.py
obucklin: This is part of a bigger conversation about joints and features. We should definitely get into this in january or february

papachap: If we go this route, `SUPPORTED_TOPOLOGY` would need to be a `list` to handle multiple connection types. That means I’d also need to update the checks in the **gh components** and the **workflow** to accommodate and process `SUPPORTED_TOPOLOGY` as a `list` instead of a single topology type.

Does everyone agree that I should go ahead with these changes? @chenkasirer  @obucklin

chenkasirer: if the current implementation of `TTenonMortise` can already handle other topologies (L basically), I'd rename it to `TenonMortise`. I'd also go ahead an make `SUPPORTED_TOPOlOGIES` a list, but it would also be ok for me if we do this in a separate PR as in most cases this is not enforced.  @obucklin what do you think?

papachap:
@chenkasirer I refactored based on what we discussed. Ready for review

src/compas_timber/design/workflow.py
```diff
@@ -164,10 +164,11 @@ def comply(self, beams, max_distance=1e-6):
...
                 solver = ConnectionSolver()
-                if (
-                    self.joint_type.SUPPORTED_TOPOLOGY
-                    == solver.find_topology(beams[0], beams[1], max_distance=max_distance)[0]
-                ):
+                found_topology = solver.find_topology(beams[0], beams[1], max_distance=max_distance)[0]
+                supported_topo = self.joint_type.SUPPORTED_TOPOLOGY
+                if not isinstance(supported_topo, list):
+                    supported_topo = [supported_topo]
+                if found_topology in supported_topo:
```
chenkasirer, line 171: we should eventually change all `SUPPORTED_TOPOLOGY` attributes so that the only check necessary is 
```python
if found_topology in self.joint_type.SUPPORTED_TOPOLOGY:
   ...
```
but we can do that in a separate PR

chenkasirer (approved):

src/compas_timber/design/workflow.py
papachap: I agree. But as you said, I thought that it would be better if it happened in a separate PR. I can create an issue in the backlog.

PR #367, opened 2025-01-15
BTLx_feat_def

obucklin:
Reworked to pass an instance of the Processing. each Processing that wants to implement this will need a `from_shapes_and_element()` classmethod with the arguments that they want to expose to the design context.

### Checklist


obucklin:
This ended up being a pretty big change. here are the main points:

- I created a `BTLxFromGeometryDefinition` class that can defer feature creation and application to a later point. 
  - This class takes a `BTLxProcessing` instance, geometry, and kwargs as inputs. 
  - This class has a method `feature_from_element(element)` that requires that the Processing type have a  `from_shapes_and_element()` constructor which is basically a wrapper for the other geometric constructor e.g. `Drilling.from_line_and_element()`.
  - This class is used 2 ways:
    - to define `BTLxProcessing`s for `FastenerTimberInterface` in which case the geometry is defined relative to the worldXY plane and applied by the joints. 
    - to define `BTLxProcessing`s to elements in the model, which is done through the features input on the model. 
- I created a `BTLxFromGeometry` which polls all BTLxProcessing child classes for the `from_shapes_and_element()` and uses that method's arguments to instantiate input parameters. 
- The `BTLxFeatureDefinition` component that previously plugged into model component features input is now `BTLxFromParams`, and applies the features directly to elements (see the Beam_From_Curve component). 

A bunch of GH components got reworked and split into different components here.

I also had to do a rework of the `DoubleCut.from_planes_and_element()` class method. 

let me know what you think, and we can gladly sit together next week. 

I will try to do unit tests by the end of the week.

PR #375, opened 2025-01-24
Text BTLxProcessing

obucklin:
New `Text` BTLx Processing.
Still needs unit tests. 
Still needs a workflow for adding the processing based on element attributes, placement, etc. 
Still needs a way to visualize geometry.

### Checklist


obucklin:
Hey Panos, lets work on this together to see if we can polish it up.

src/compas_timber/fabrication/basic_characters.json
chenkasirer, line 1: what's this file for?

src/compas_timber/ghpython/components/CT_Label_Beams/code.py
```diff
@@ -0,0 +1,35 @@
+from compas.scene import Scene
+from ghpythonlib.componentbase import executingcomponent as component
+
+from compas_timber.fabrication import Text
+
+
+class MyComponent(component):
```
chenkasirer, line 7: please give meaningful name

src/compas_timber/ghpython/components/CT_Label_Beams/code.py
```diff
@@ -0,0 +1,35 @@
...
+    def RunScript(self, model, base_string=None, char_to_replace=None, attributes=None, text_size=None):
+        scene = Scene()
+        self.clear_labels(model)
+        for element in model.beams:
+            for feature in element.features:
+                if getattr(feature, "is_label", None):
+                    element.remove_features(feature)
```
chenkasirer, line 14: aren't these two doing the same?

src/compas_timber/ghpython/components/CT_Label_Beams/code.py
```diff
@@ -0,0 +1,35 @@
...
+from compas_timber.fabrication import Text
+
+
+class MyComponent(component):
+    def RunScript(self, model, base_string=None, char_to_replace=None, attributes=None, text_size=None):
+        scene = Scene()
+        self.clear_labels(model)
+        for element in model.beams:
+            for feature in element.features:
+                if getattr(feature, "is_label", None):
```
chenkasirer, line 13: I find it better to check `feature.PROCESSING_NAME == "Text"` here then to extend the public interface of just that one process with `is_label`

src/compas_timber/ghpython/components/CT_Label_Beams/metadata.json
```diff
@@ -0,0 +1,54 @@
+{
```
chenkasirer, line 1: copy paste accident? ;)

chenkasirer (changes requested):

src/compas_timber/fabrication/basic_characters.json
obucklin: This contains the polylines to draw the text characters

src/compas_timber/ghpython/components/CT_Label_Beams/code.py
obucklin: The idea was to leave `Text` processings that were generated elsewhere. I tried to do this with a persistent list in the component, but I couldn't figure that out. The problem is that this modifies the elements in the original model, so if any parameters change, you end up with multiple labels. I see a few options:
-Copy the model in this component and add the labels to the new elements.
-give the elements a `label` attribute which is a single `Text` processing. this would be updated by the `LabelBeams` component.
-give the processings a `is_label` boolean, allowing us to identify and remove them.

@chenkasirer what do you think? any other options?

src/compas_timber/ghpython/components/CT_Label_Beams/code.py
obucklin: see comment above

src/compas_timber/ghpython/components/CT_Label_Beams/code.py
obucklin: > The idea was to leave `Text` processings that were generated elsewhere. I tried to do this with a persistent list in the component, but I couldn't figure that out. The problem is that this modifies the elements in the original model, so if any parameters change, you end up with multiple labels. I see a few options: -Copy the model in this component and add the labels to the new elements. -give the elements a `label` attribute which is a single `Text` processing. this would be updated by the `LabelBeams` component. -give the processings a `is_label` boolean, allowing us to identify and remove them.
> 
> @chenkasirer what do you think? any other options?

@chenkasirer you have any thoughts about how to do this?

chenkasirer: mmh my I like the idea of a "LabelBeams" component where you tell it which attribute of of the beam you want and it takes and uses it as the text. If the component spits out a `BTLxFromGeometryDefinition`, then should fit nicely within the current system.. or am I missing something? 

Actually setting the attribute could be done as part of the BeamFromCenterline componet or outside the scope of this (AKA should already be set like e.g. the key)

src/compas_timber/fabrication/basic_characters.json
chenkasirer: it rather large. maybe try to save it as not prettified (not indentations) or even take a look at: https://compas.dev/compas/latest/api/generated/compas.data.json_dumpz.html

src/compas_timber/ghpython/components/CT_Label_Beams/code.py
obucklin: The idea was that this label also responds to attributes generated at the `model` phase and would include attributes like `group`, assembly order, etc. I guess technically that stuff is already in the element component, but in terms of workflow it makes sense to have this post-process the `model.elements`.

obucklin:
Hey @papachap @chenkasirer,  I reworked this to have a better GH workflow and to hopefully permit more robust expansion later. I essentially did the following:

Added a `Label` class in `compas_timber.planning`. This allows us to define the content of the Label (the text, element attributes, etc) separately, which we can then pass to different label types, like a `Text` `BTLxProcessing`, but also to normal printed labels or stickers. This could be expanded to make human and machine-readable labels. 

`Label` also has a function to find the biggest free gap on the beam and to place the `Text` processing there. 

There is a `DeferredLabel` which allows us to pass the label to the model GH component.  This uses the same `feature_from_elements` method to get the `BTLxProcessing` allowing us to use it in the same way as the other feature defs.

I will work on some unit tests for the `Text` and `Label`, but if you have time, it would be great to get some feedback. have a nice weekend!

obucklin:
added unit tests. ready for review

src/compas_timber/fabrication/text.py
```diff
@@ -0,0 +1,285 @@
...
+                pt.transform(Transformation.from_frame_to_frame(Frame.worldXY(), face))
+        return string_curves
+
+
+class TextParams(BTLxProcessingParams):
+    def __init__(self, instance):
+        super(TextParams, self).__init__(instance)
+
+    def as_dict(self):
+        result = super(TextParams, self).as_dict()
```
chenkasirer, line 274: here also, align with `main`

src/compas_timber/fabrication/text.py
```diff
@@ -0,0 +1,285 @@
...
+
+        Returns
+        -------
+        :class:`compas.geometry.Brep`
+            The resulting geometry after processing. #TODO: think about ways to display text curves from `draw_string_on_element()`
+
+        """
+        return geometry
+
+    def draw_string_on_element(self, element):
```
chenkasirer, line 229: I understand we're not doing this at the moment but we might want to in the future, right?
could you add a docstring so it's clearer what this does?

src/compas_timber/planning/label.py
```diff
@@ -0,0 +1,157 @@
...
+                string_out += char
+            else:
+                string_out += str(getattr(element, att_names.pop(0), " ") if len(att_names) > 0 else "_")
+        for attribute in att_names:
+            string_out += "_"
+            string_out += str(getattr(element, attribute, " "))
+
+        return cls(element, string_out)
+
+    def engrave_on_beam(self, beam, text_height=None, ref_side_index=0):
```
chenkasirer, line 67: should this maybe be an alternative constructor of the `Text`processing? (e.g. `Text.from_element_and_text`)

src/compas_timber/planning/label.py
```diff
@@ -0,0 +1,157 @@
...
+        return {"element": self.element, "text": self.text}
+
+    @property
+    def text(self):
+        if self.user_text:
+            return self.user_text
+        return "{}_{}_{}".format(self.element.__class__.__name__, self.element.key, self.element.attributes.get("category") if self.element.attributes.get("category") else "")
+
+    @classmethod
+    def from_element(cls, element, attributes=[], base_string="", char_to_replace=None):
```
chenkasirer, line 33: immutable objects (`[]`) should not be set as default value as these are only evaluated once

src/compas_timber/planning/__init__.py
```diff
@@ -18,4 +20,6 @@
     "Step",
     "SimpleSequenceGenerator",
     "Text3d",
+    "Label",
+    "DeferredLabel",
```
chenkasirer, line 24: dunno if that's the based place for these, also the BuildingPlan stuff might end up moving to compas XR.
how about the `workflow.py` module? is that not a better fit?

src/compas_timber/planning/label.py
```diff
@@ -0,0 +1,157 @@
...
+    text : str
+        The text of the label.
+    """
+
+    def __init__(self, element, text=None):
+        self.element = element
+        self.user_text = text
+
+    @property
+    def __data__(self):
```
chenkasirer, line 23: we only need the data interface if this needs to be serialized. since it doesn't end on any beam maybe it doesn't. 
if it does, then it should inherit from `Data`

src/compas_timber/planning/label.py
```diff
@@ -0,0 +1,157 @@
...
+        self.attributes = attributes
+        self.char_to_replace = char_to_replace
+        self.text_height = text_height
+        self.ref_side_index = ref_side_index
+
+    def __str__(self):
+        return "DeferredLabel: {}".format(self.base_string)
+
+    @property
+    def __data__(self):
```
chenkasirer, line 142: same here, not sure if this needs `Data`

src/compas_timber/planning/__init__.py
obucklin: I don't have a real opinion on this, except that workflow is maybe becoming a bit of a catchall. I would leave this as its own module. Label is perhaps most relevant for fabrication, or maybe we need to consider an `assembly` package (for assembly processes).

src/compas_timber/planning/label.py
obucklin: I'm still not sure on when we need to be able to serialize, and how to best do it. Shouldn't every compas object be serializable? When should a class inherit from `object` vs `Data`?

src/compas_timber/planning/label.py
obucklin: see my above response

src/compas_timber/planning/label.py
obucklin: I see this as essentially like the `Joint.add_features()` method. I could add an alternate constructor to `Text` but I would keep this method here BC I think we want the option to call `Label.print_label` or `Label.mark_element` (yes some timber CNCs have marker tools)

obucklin:
@chenkasirer not ready for review, but I have some questions and comments of my own.

PR #390, opened 2025-02-11
fix `max_distance` bugs

obucklin:
Bug fixes after adding `max_distance` to joint rules. 

### Checklist

### DIFF
--- a/src/compas_timber/design/workflow.py
+++ b/src/compas_timber/design/workflow.py
@@ -91,23 +91,24 @@ class JointRule(object):
         element_pairs = solver.find_intersecting_pairs(elements, rtree=True, max_distance=max_rule_distance)
         joint_defs = []
         unmatched_pairs = []
+        compliant_direct_rules = []
+        for rule in direct_rules:
+            if rule.comply(rule.elements, model_max_distance=max_distance):  # see if pair complies with max distance
+                joint_defs.append(JointDefinition(rule.joint_type, rule.elements, **rule.kwargs))
+                compliant_direct_rules.append(rule)
         while element_pairs:
             pair = element_pairs.pop()
             match_found = False
-            for rule in direct_rules:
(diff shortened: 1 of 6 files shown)


src/compas_timber/utils/__init__.py
```diff
@@ -190,6 +190,12 @@ def distance_segment_segment(segment_a, segment_b):
 
     """
     pta, ptb = intersection_line_line(segment_a, segment_b)
+    if not pta:  # segments are parallel
+        dists = []
+        for pair in [(segment_a[0], segment_b[0]), (segment_a[1], segment_b[0]), (segment_b[0], segment_a[1]), (segment_b[1], segment_a[1])]:
+            """gets shortest distance between all 4 possible pairs of endpoints. only for L/I_Topo. T_Topology cannot have paralell segments"""
```
chenkasirer, line 196: please use `#` for in-code comments

src/compas_timber/design/workflow.py
```diff
@@ -91,23 +91,24 @@ def joints_from_beams_and_rules(elements, rules, max_distance=TOL.absolute):
         element_pairs = solver.find_intersecting_pairs(elements, rtree=True, max_distance=max_rule_distance)
         joint_defs = []
         unmatched_pairs = []
+        compliant_direct_rules = []
+        for rule in direct_rules:
+            if rule.comply(rule.elements, model_max_distance=max_distance):  # see if pair complies with max distance
```
chenkasirer, line 96: if rule already contains the elements, why does do they need to be passed from the outside?

src/compas_timber/utils/__init__.py
chenkasirer: if possible, try to use something from `https://docs.python.org/3/library/itertools.html` here instead. would make for a more readable code.

chenkasirer (approved):
lgtm

PR #391, opened 2025-02-12
Processing proxy

chenkasirer:
As noticed by @jonashaldemann, the new processing system tends to be quite a bit slower then the previous one. This is a little experiment trying to get some improvement.

Implemented a proxy object for `JackRafterCut` that's a drop-in replacement and behaves almost like it.
`JackRafterCutProxy` can be created similarly to `JackRafterCut` but it does not calculate machining parameters until they are requested either by calling `unproxified()` (which returns an actual instance of `JackRafterCut`) or by trying to call any attributes that are not available in `JackRafterCutProxy`. such calls are then passed through to an instance of `JackRafterCut`.

I replaced `JackRafterCut` with the proxy version in some joints and noticed some improvement.

We can implement these for some other processing or not depending on-demand, as otherwise everything behaves the same.

incidentally this creates a deferring effect which I know @obucklin you're working on, perhaps these could  be eventually merged as they are very similar.

### Checklist


jonashaldemann:
looks good to me.

jonashaldemann (approved):

PR #399, opened 2025-02-18
added component, ready for PR

obucklin:
This is my solution for a deferred application of `BTLx_from_Params`. I put the `DeferredBTLxProcessing` directly in the component...

### Checklist


src/compas_timber/ghpython/components/CT_BTLx_From_Params/code.py
```diff
@@ -0,0 +1,79 @@
...
+        super(BTLxFromParams, self).__init__()
+        self.classes = {}
+        for cls in get_leaf_subclasses(BTLxProcessing):
+            self.classes[cls.__name__] = cls
+        if ghenv.Component.Params.Output[0].NickName == "Processing":
+            self.processing_type = None
+        else:
+            self.processing_type = self.classes.get(ghenv.Component.Params.Output[0].NickName, None)
+
+    def RunScript(self, element, ref_side=0, *args):
```
chenkasirer, line 24: the default ref side here doesn't seem to work, I'm getting a nasty error when not setting a ref_side

src/compas_timber/ghpython/components/CT_BTLx_From_Params/code.py
```diff
@@ -0,0 +1,79 @@
...
+    def __str__(self):
+        return "DeferredBTLxProcessing({})".format(self.processing.__class__.__name__)
+
+    def __repr__(self):
+        return "DeferredBTLxProcessing(processing={!r})".format(self.processing.__class__.__name__)
+
+    def ToString(self):
+        return self.__str__()
+
+    def feature_from_element(self, element):
```
chenkasirer, line 78: `element` is given but not used?

chenkasirer (commented):
@jonashaldemann @Stryzhevska perhaps it's worth documenting it well that, when combined with joinery, this preview feature might result in processing that looks different in the preview and in the final geometry coming back from model. For example, this is the same drilling appearing in different places due to the beam getting extended cause of the miter.

image

src/compas_timber/ghpython/components/CT_BTLx_From_Params/code.py
obucklin: so that the signature matches the other deferred feature type (`BTLxFromGeometryDefinition`)

obucklin:
> @jonashaldemann perhaps it's worth documenting it well that, when combined with joinery, this preview feature might result in processing that looks different in the preview and in the final geometry coming back from model. For example, this is the same drilling appearing in different places due to the beam getting extended cause of the miter.
> 
> image

I put in the hint of the geometry output that this was likely to change. We could just have a persistent warning saying stating that fact, leaving the component always orange...

PR #403, opened 2025-02-24
Fix `YButtJoint` bug

papachap:
This PR fixes a bug in the `YButtJoint` and enhances error handling in `DoubleCut` and `TimberModel.process_joinery()`. 

In detail:
- Refactored `YButtJoint.check_element_compatibility()` to validate the coplanarity and dimensions of cross elements.
- Updated `DoubleCut.from_planes_and_beam()` to check for parallel planes and raise a `ValueError` when detected.
- Modified `process_joinery` to catch `ValueError` exceptions during `BTLxProcessing` generation and wrap them in `BeamJoiningError` objects.

### Checklist


src/compas_timber/fabrication/double_cut.py
```diff
@@ -199,7 +199,12 @@ def from_planes_and_beam(cls, planes, beam, ref_side_index=None):
 
         # convert all frames to planes
         planes = [Plane.from_frame(plane) if isinstance(plane, Frame) else plane for plane in planes]
-        """ get the intersection line of cutting planes, which is used to determine whether the planes cut a concave or convex shape"""
+
+        # check if the planes are parallel
+        if TOL.is_close(dot_vectors(planes[0].normal, planes[1].normal), 1.0):
```
obucklin, line 204: Why should the planes not be parallel? I guess that would be better to cut with a `JackRafterCut` but then we have to make that substitution here.

obucklin (changes requested):

src/compas_timber/fabrication/double_cut.py
papachap: i added this part, because if the planes are parallel the intersection_plane_plane function returns none. The error is then not caught and makes it difficult to debug. 

It was a quick fix for something i noticed while refactoring something else. In case we want a more comprehensive solution, i would suggest we do it on a different PR.

obucklin (approved):
The longest Bug Fix ever. LGTM

src/compas_timber/connections/utilities.py
```diff
@@ -146,7 +148,7 @@ def beam_ref_side_incidence_with_vector(beam_a, vector, ignore_ends=True):
     return ref_side_angles
 
 
-def are_beams_coplanar(beam_a, beam_b, tol=TOL):
+def are_beam_frames_orthogonal(beam_a, beam_b, tol=TOL):
```
chenkasirer, line 151: mmh orthogonal to me means perpendicular.. is this what this function checks? because given the implementation (and the docstring), it's co-planarity that's actually checked, right?

obucklin: yeah I guess you are correct about the meaning of orthogonal. This function returns true if the beams are either parallel or perpendicular... I think `are_beams_coplanar` was definitely too vague, but I'm open to other name suggestions.

obucklin: > In math, this adjective refers to geometric lines or planes that are not parallel or perpendicular to a line or surface.

use `are_beam_frames_oblique` ?

chenkasirer: mmh I see. well this sound to me like perhaps this needs to be split into two separate functions. 

client code can simply go:
```python
are_beams_perpendicular(beam_a, beam_b) or are_beams_parallel(beam_a, beam_b)
```
that'll also make them reusable independently of each other

obucklin: I realized this function has an different purpose than to tell if the frames are not oblique. I changed the name accordingly to `are_beams_aligned_with_cross_vector`

PR #407, opened 2025-02-26
y-joint fix

obucklin:
changed it to generate a JackRafterCut when the cross beams are parallel.

### Checklist


src/compas_timber/connections/utilities.py
```diff
@@ -165,6 +167,13 @@ def are_beams_coplanar(beam_a, beam_b, tol=TOL):
         True if the beams are coplanar, False otherwise.
     """
     # Compute the cross product of the centerline directions of the two beams
+    beam_angle = angle_vectors(beam_a.centerline.direction, beam_b.centerline.direction)
+    if TOL.is_zero(beam_angle) or TOL.is_zero(beam_angle - math.pi):  # beams are parallel
+        angle = angle_vectors(beam_a.centerline.direction, beam_b.centerline.direction)
+        if TOL.is_zero(angle) or TOL.is_zero(angle - math.pi):
```
chenkasirer, line 173: is this not the same check twice?

src/compas_timber/connections/utilities.py
```diff
@@ -165,6 +167,13 @@ def are_beams_coplanar(beam_a, beam_b, tol=TOL):
...
+            if TOL.is_close(beam_a.width, beam_b.width) and TOL.is_close(beam_a.height, beam_b.height):
```
papachap, line 174: I have a feeling this won't always work, since the width and the height of a beam are relative to the it's normal. Meaning that there could be cases where the expected matching dimensions are the width of one beam and the height of the other. 

Instead you can check the `get_dimensions_relative_to_side()` method in `Beam`.

papachap: FIY this is already checked in the `check_elements_compatibility()` method of the `YButtJoint`, so there's no need to check it again.

papachap (changes requested):

src/compas_timber/connections/utilities.py
obucklin: I was getting a division by zero error, so I did it this way.

src/compas_timber/connections/utilities.py
obucklin: I made some changes to this, and it checks coplanarity in the direction of the cross_vector. I changed the check in `YButtJoint` to only check the height in the direction of the main beam using `get_dimensions_relative_to_side`

obucklin (commented):
I don't understand how git works. I thought @papachap you were supposed to review my changes and then I could merge them into your branch and then I could review your changes and you could merge them into main...

src/compas_timber/connections/utilities.py
```diff
@@ -165,16 +167,38 @@ def are_beams_coplanar(beam_a, beam_b, tol=TOL):
...
+        # both beam z-vectors perpendicular to cross vector
+        if beam_a.width == beam_b.width:
+            return True
+    elif tol.is_close(dot_with_beam_a_normal, 1.0) and tol.is_zero(dot_with_beam_b_normal):
+        if beam_a.height == beam_b.width:
+            return True
+    elif tol.is_zero(dot_with_beam_a_normal) and tol.is_close(dot_with_beam_b_normal, 1.0):
+        if beam_a.width == beam_b.height:
+            return
+    return False
```
papachap, line 201: I'm still hesitant about this change. As I see it, co-planarity is independent of the beam dimensions.

For example, in the case of the `TStepJoint`, the beams must be co-planar, but their section dimensions are irrelevant (which btw I just realized I'm duplicating the function's code instead of calling it).

The way I see it, you should enhance the function to check properly for co-planarity if the beams are co-linear (ie `if cross_vector.length == 0.0`), and do the dimensions check elsewhere (which you already do in the `check_elements_compatibility()` method of the `YButtJoint`).

obucklin: Ok I basically reverted it but kept the check for parallel beams, in which case it just compares the two beam.frame.normal voctors directly. Also remamed it for clarity.

obucklin:
ok I made the changes we discussed.

papachap (approved):
LGTM!

PR #411, opened 2025-03-04
Lap_DS_OB

obucklin:
My solution to getting the pocket params. I am getting <0.1mm imprecision now. It owuld be good to know if there is any improvement on the recip. frame model from CiA. I am somewhat convinced that the remaining inprecision is because of the inherent 2-3 decimal precision in BTLx and the fact that each processing is defined in local part coordinates. 

- [x] Maybe a bug fix in a **backwards-compatible** manner, could also be the same as before.

### Checklist


papachap:
image
image

@obucklin unfortunately the result is exactly the same as in the current state of the initial `Pocket` PR (pushed some changes yestarday). 
In the images above you can see an `XLapJoint` from the latest design iteration I got from Lorin, generated from the two github branches. It's the one with the biggest offset i could spot, and according to the measurement in the BTL viewer it is almost 1mm..

PR #420, opened 2025-03-12
Bug Fix in `shape_strings`

papachap:
Fixed a bug occurring when generating the shape polyline, by ensuring the polyline is always closed.

### Checklist

### DIFF
--- a/src/compas_timber/utils/__init__.py
+++ b/src/compas_timber/utils/__init__.py
@@ -292,7 +292,7 @@ def distance_segment_segment(segment_a, segment_b):
 
 
 def is_polyline_clockwise(polyline, normal_vector):
-    """Check if a polyline is clockwise.
+    """Check if a polyline is clockwise. If the polyline is open, it is closed before the check.
 
     Parameters
     ----------
@@ -305,7 +305,10 @@ def is_polyline_clockwise(polyline, normal_vector):
         True if the polyline is clockwise, False otherwise.
 
     """
-    assert polyline[0] == polyline[-1], "Polyline should be closed"
+    # make sure the polyline is closed
+    if not polyline[0] == polyline[-1]:
+        polyline.append(polyline[0])
+
     angle_sum = 0
     for i in range(len(polyline) - 1):
         u = Vector.from_start_end(polyline[i - 1], polyline[i])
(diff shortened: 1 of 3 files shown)


obucklin (dismissed):
Hey the lists of points you are changing should not have closed points. If you look at sample BTLx files, they don't have the first and last point index as the same, aka a 4-sided BrepFaceLoop is represented with 4 point indices in BTLx. this will also break the check on line 451/453, which avoids making surfaces with only 2 points, which prevents faces from being formed that are zero-area/ just a line. 

What is the situation that is causing a problem for you? I expect there is some other cause.

papachap:
> Hey the lists of points you are changing should not have closed points. If you look at sample BTLx files, they don't have the first and last point index as the same, aka a 4-sided BrepFaceLoop is represented with 4 point indices in BTLx. this will also break the check on line 451/453, which avoids making surfaces with only 2 points, which prevents faces from being formed that are zero-area/ just a line.
> 
> What is the situation that is causing a problem for you? I expect there is some other cause.

In `correct_polyline_direction` and `is_polyline_clockwise()`, an assertion checks whether the first and last points of the polyline are the same. If they are not (which based on what you say it should always be the case), an assertion error is raised.
Apparently I have managed to export a btlx where the shapes seem reasonable, without any issues being raised while doing so.

src/compas_timber/utils/__init__.py
```diff
@@ -305,7 +305,10 @@ def is_polyline_clockwise(polyline, normal_vector):
         True if the polyline is clockwise, False otherwise.
 
     """
-    assert polyline[0] == polyline[-1], "Polyline should be closed"
+    # make sure the polyline is closed
+    if not polyline[0] == polyline[-1]:
+        polyline.append(polyline[0])
```
chenkasirer, line 310: As a client, I wouldn't expect a function called `is_polyline_clockwise` to modify what I give it.
if polyline has to be closed for the check, but the function should support both closed and opened polylines then make the check on a copy.

chenkasirer (approved):
LGTM!

chenkasirer (approved):

chenkasirer:
ah @obucklin needs to approve as he asked for changes

papachap:
@obucklin approve?

PR #424, opened 2025-04-09
Timber Plate with inclined edges

obucklin:
This is a big change of the `Plate` element and a rework of the `FreeContour` processing. The Plate is now defined by top and bottom closed polylines, instead of a  polyline and thickness. 

The `FreeContour` processing can now generate inclined cuts. This can be defined for the whole processing or on a per segment basis. The contour can generate the inclination itself via the `from_polylines_and_element` constructor, or can take the inclination in the normal `__init__()` constructor. 

`FreeContour` can now generate `DualContour` processings where the cut is defined by a top and bottom contour. `FreeContour` will generate a normal `btlx.Contour` object if each segment of the associated contour is parallel to its principal contour counterpart. If not, then a `DualContour` is generated. 

The `FreeContour` constructors can now identify which `ref_side_index` they should be applied to.

### Checklist


src/compas_timber/elements/plate.py
```diff
@@ -206,6 +230,20 @@ def add_feature(self, feature):
...
+        # type: () -> compas.geometry.Brep
+        """The shape of the plate before other features area applied.
+
+        Returns
+        -------
+        :class:`~compas.geometry.Brep`
+            The shape of the element.
+
+        """
+        plate_geo = Brep.from_loft([NurbsCurve.from_points(pts, degree=1) for pts in [self.outline_a, self.outline_b]])
```
chenkasirer, line 243: opt for tuple where collection is anyways immutable
```suggestion
        plate_geo = Brep.from_loft([NurbsCurve.from_points(pts, degree=1) for pts in (self.outline_a, self.outline_b)])
```

src/compas_timber/elements/plate.py
```diff
@@ -1,43 +1,58 @@
...
-    outline : :class:`~compas.geometry.RhinoCurve`
-        A line representing the outline of this plate.
-    thickness : float
-        Thickness of the plate material.
-    vector : :class:`~compas.geometry.Vector`, optional
-        The vector of the plate. Default is None.
+    outline_a : :class:`~compas.geometry.RhinoCurve`
+        A line representing the principal outline of this plate.
+    outline_b : :class:`~compas.geometry.RhinoCurve`
+        A line representing the associated outline of this plate. This should have the same number of points as outline_a.
```
chenkasirer, line 27: these should be Polylines, right?

src/compas_timber/elements/plate.py
```diff
@@ -178,17 +192,27 @@ def key(self):
...
-    def vector(self):
-        if not self._vector:
-            self._vector = self.frame.zaxis * self.thickness
-        return self._vector
+    # ==========================================================================
+    # Alternate constructors
+    # ==========================================================================
+
+    @classmethod
+    def from_outline_thickness(cls, outline, thickness, vector=None, frame=None, blank_extension=0, **kwargs):
```
chenkasirer, line 206: please add docstring

chenkasirer: also, is it really necessary to accept `vector` and `frame`? seems to be introducing a lot of complexity to the body of the method.

src/compas_timber/elements/plate.py
```diff
@@ -178,17 +192,27 @@ def key(self):
...
+        thickness_vector = Frame.from_points(outline[0], outline[1], outline[-2]).normal
+        if vector and thickness_vector.dot(vector) < 0:
+            thickness_vector = -thickness_vector
+        elif frame and thickness_vector.dot(frame.zaxis) < 0:
+            thickness_vector = -thickness_vector
+        thickness_vector.unitize()
+        thickness_vector *= thickness
+        outline_b = Polyline([p.translated(thickness_vector) for p in outline])
```
chenkasirer, line 214: why not just `outline.translated(thickness_vector)`?

src/compas_timber/elements/plate.py
```diff
@@ -246,9 +285,9 @@ def compute_aabb(self, inflate=0.0):
             The AABB of the element.
 
         """
-        vertices = [point for point in self.outline.points]
-        for point in self.outline.points:
-            vertices.append(point + self.vector)
+        vertices = []
+        for outline in [self.outline_a, self.outline_b]:
```
chenkasirer, line 289:
```suggestion
        for outline in (self.outline_a, self.outline_b):
```

src/compas_timber/fabrication/btlx.py
```diff
@@ -251,7 +251,7 @@ def _create_processing(self, processing):
         return processing_element
 
     def _element_from_complex_param(self, param):
-        serializer = self.SERIALIZERS.get(type(param), None)
+        serializer = self.SERIALIZERS.get(type(param).__name__, None)
```
chenkasirer, line 254: why use name instead of type?

src/compas_timber/fabrication/free_contour.py
```diff
@@ -11,40 +19,58 @@
...
+    inclination : float or list of float, optional
+        The inclination of the contour. Default is 0 meaning the cut is perpendicular to the reference side.
 
     """
 
     PROCESSING_NAME = "FreeContour"  # type: ignore
 
-    def __init__(self, contour_points, depth, counter_sink=False, tool_position=AlignmentType.LEFT, depth_bounded=False, inclination=0, **kwargs):
+    def __init__(
+        self, contour_points, associated_contour_points=None, depth=None, counter_sink=False, tool_position=AlignmentType.LEFT, depth_bounded=True, inclination=None, **kwargs
```
chenkasirer, line 52: groups of arguments that are mutually exclusive and attributes which accept multiple types are a sign to me that we should either:
- split this class into 2 classes, e.g. `FreeContour` and `FreeDualContour`
or
- have this class accept either `Contour` or `DualContour` instance (aka dependency injection), we could then make factory methods which help the user create either one.

I think I prefer the 2nd option, but we can maybe discuss this and agree on one together.

src/compas_timber/elements/plate.py
obucklin: for the moment. It would be good to move towards NurbsCurve evtl.

src/compas_timber/elements/plate.py
obucklin: I don't know if the frame is necessary. The frame is generated from the `Outline` by default, but there is no clear location for the frame the way there is with a `Beam`. Do you think it is important that the user can define where they want the frame to be? I can also see times when they may want to define the frame and vector differently, for example on sheeting for a wall element, where you might want to keep the frames aligned, but one offsets to the inside and one to the outside. I can remove if you think it's better.

src/compas_timber/fabrication/btlx.py
obucklin: type was not returning the serializer method, so I did this.

src/compas_timber/elements/plate.py
obucklin: I tried that and it didn't work...

obucklin: Ok I had to cast to/reinstaintiate Polyline but it works

src/compas_timber/fabrication/free_contour.py
obucklin: I was thinking about this a bit. There is not a one-to-one relationship between 2 curve inputs and a `DualContour` BTLx type. In the case where the top and bottom polyline inputs have parallel corresponding segments, it still generates a normal `Contour`.

src/compas_timber/elements/plate.py
obucklin: separate Frame also would allow the user to define the orientation of the blank. may be useful for nesting or dealing with sheet material that has directionality

CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Refactored `_create_negative_volumes()` in `LapJoint` so that it generates box-like volumes. 
 * Refactored `XLapJoint`, `LLapJoint`, `TLapJoint` so that they use the `_create_negative_volumes()` method to get the negative volumes and use the alt constructor `Lap.from_volume_and_beam()`.
 * Fixed an error occuring in `BTLxPart.shape_strings` by ensuring the polyline is always closed.
+* Implemented `Inclination` in the `FreeContour` BTLx Processing.
+* Changed `Plate` element to be defined by top and bottom contours instead of one contour and thickness.
```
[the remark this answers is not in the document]
obucklin: added to `### Added`

src/compas_timber/fabrication/free_contour.py
```diff
@@ -11,40 +19,37 @@
...
+        The contour parameter object.
     counter_sink : bool, optional
         If True, the contour is a counter sink. Default is False.
     tool_position : str, optional
         The position of the tool. Default is "left".
     depth_bounded : bool, optional
         If True, the depth is bounded. Default is False, meaning the machining will cut all the way through the element.
-    inclination : float, optional
-        The inclination of the contour. Default is 0. This is not yet implemented.
+    inclination : float or list of float, optional
```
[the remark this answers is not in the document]
obucklin: yeah it got moved to the `contour_param_object`

src/compas_timber/ghpython/components/CT_Plate/code.py
```diff
@@ -4,69 +4,81 @@
...
 
 class Plate(component):
-    def RunScript(self, outline, thickness, vector, category, updateRefObj):
-        # minimum inputs required
+    def RunScript(self, outline, outline_b, thickness, vector, category, updateRefObj):
         if not outline:
-            self.AddRuntimeMessage(Warning, "Input parameter 'Outline' failed to collect data")
-        if not thickness:
-            self.AddRuntimeMessage(Warning, "Input parameter 'Thickness' failed to collect data")
+            print("Input parameter 'Outline' failed to collect data")
```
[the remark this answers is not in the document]
obucklin: yup or even this helper method.

chenkasirer (approved):
LGTM

PR #432, opened 2025-05-06
Model tolerance

chenkasirer:
* Added `tolerance` setting to `TimberModel`
* Added scaling methods to the various processings so that design in meters can be scaled when writing BTLx.
* Added a couple of missing Data implementations in processings.

### Checklist


src/compas_timber/fabrication/btlx.py
```diff
@@ -183,8 +186,11 @@ def _create_part(self, element, order_num):
             The part element.
 
         """
+        assert self._tolerance
         # create part element
-        part = BTLxPart(element, order_num=order_num)
+        scale_factor = 1000.0 if self._tolerance.unit == "M" else 1.0
```
obucklin, line 191: should we accommodate other units? cm?

src/compas_timber/fabrication/btlx.py
```diff
@@ -306,16 +318,17 @@ class BTLxPart(object):
 
     """
 
-    def __init__(self, element, order_num):
+    def __init__(self, element, order_num, scale_factor=1.0):
```
obucklin, line 321: should we make `orde_num` optional with a default value?

papachap, line 321: can you update the docstring? I see it's missing also the order_num

src/compas_timber/fabrication/btlx.py
papachap, line 191: I know designing in centimeters isn't common practice, but since we're already supporting meters as a unit, should we also consider allowing centimeters?

src/compas_timber/fabrication/btlx.py
```diff
@@ -217,6 +223,12 @@ def _create_processing(self, processing):
             The processing element.
 
         """
+        assert self._tolerance
+        # BTLx always uses mm
+        if self._tolerance.unit == "M":
```
papachap, line 228: same as above

src/compas_timber/ghpython/components_cpython/CT_Model/code.py
```diff
@@ -52,7 +54,16 @@ def RunScript(
         # clear Nones
         Containers = [c for c in Containers if c is not None]
 
-        Model = TimberModel()
+        units = Rhino.RhinoDoc.ActiveDoc.GetUnitSystemName(True, True, True, True)
+        tol = None
+        if units == "m":
```
obucklin, line 59: FYI python 3.10 and up have pattern matching/ `switch: case` function. what version of python does rhino8 GH use?

tests/compas_timber/test_pocket.py
```diff
@@ -268,3 +268,48 @@ def test_pocket_with_5_faces(beam):
     except Exception as e:
         assert isinstance(e, ValueError)
         assert "Volume must have 6 faces." in str(e)
+
+
+def test_pocket_scaled():
+    machining_limits = {
```
papachap, line 274: FYI there's a `MachiningLimits` config class that already defines the expected keys and defaults for the machining limits.

`machining_limits = MachiningLimits().limits`

obucklin (approved):
This is a very cool and ambitions change. I am wondering how this will change when considering non-BTLx translations for example Revit(m) to cadwork(mm). I think this is a good way to start and then we can cross that bridge when we get there.

papachap (approved):
LGTM! left a couple of comments

src/compas_timber/fabrication/btlx.py
chenkasirer: we could, but perhaps in a separate PR. currently the only thing that makes it in to `order_num` is anyways the index of the element in the model. there's no way for user to specify anything else as this is internal writer logic, so even if we made it optional, it will not change the behavior.

src/compas_timber/ghpython/components_cpython/CT_Model/code.py
chenkasirer: `3.9` in Rhino8 ;/ but Rhino9 will be shipped with `3.13`! https://discourse.mcneel.com/t/python-3-13-in-rhino-wip/205209

src/compas_timber/fabrication/btlx.py
chenkasirer: yup, similar comment from @obucklin. will check out what's up with `Tolerance`, currently can't specify anything other than mm and meters there.

tests/compas_timber/test_pocket.py
chenkasirer: nice!

PR #435, opened 2025-05-07
Update header

chenkasirer:
* Added task to update the minimum version header in all cpython components.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Added tasks `update-gh-header` to update the version in the header of the GH components.
+
 ### Changed
 
 ### Removed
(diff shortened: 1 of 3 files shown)


chenkasirer:
> > Hey I don't know much about how this all works, but I trust that it does. It would be cool to get a quick rundown of the componentizer workflow at some point.
> 
> Same here, maybe when @chenkasirer is back he can walk us through.

Sure thing. let's sit and I walk you through it.

In short though, each of our Rhino8 components has a header that specifies the minimum version of compas timber required to run it. it looks like that:
```
# r: compas_timber>=0.15.3
```
Say we add something that's only available in version `0.16.2`, and use it in one of our components, the next GH package release should specify `0.16.2` as the minimum version, otherwise our component will fail.
Since this header is present in every single component, we'd need to change them one by one. so this PR adds a task in `tasks.py` that automates that. 

```batch
> invoke -h update-gh-header
Usage: inv[oke] [--core-opts] update-gh-header [--options] [other tasks here ...]

Docstring:
  Update the minimum version header of all CPython Grasshopper components.

Options:
  -v STRING, --version=STRING   New minimum version to set in the header. If not provided, current version is used.
```

PR #436, opened 2025-05-08
Processings data

chenkasirer:
* Serialize features that are not joinery related
this is done by an added flag `is_joinery` to the `BTLxProcessing`s
* Moved feature management to the `TimberElement` parent class
* Fixed some visualization bug in `Plate`.

### Checklist

### DIFF
--- a/src/compas_timber/elements/plate.py
+++ b/src/compas_timber/elements/plate.py
@@ -83,6 +83,9 @@ class Plate(TimberElement):
         self.debug_info = []
         self._ref_frame = None
         self._blank = None
+
+        # The base feature is expected to be the first as it's treated differently than the others
+        # therefore, if there are any feature passed to __init__, they should appear after the base feature
         self._features.insert(0, FreeContour.from_top_bottom_and_elements(self.outline_a, self.outline_b, self, interior=False))
 
     def __repr__(self):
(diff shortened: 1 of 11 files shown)


src/compas_timber/elements/plate.py
```diff
@@ -265,6 +245,8 @@ def shape(self):
         """
         plate_geo = Brep.from_loft([NurbsCurve.from_points(pts, degree=1) for pts in (self.outline_a, self.outline_b)])
         plate_geo.cap_planar_holes()
+        if plate_geo.volume < 0:
```
obucklin, line 248: This may be a case for `compas.geometry.BrepOrientation`. maybe not implemented for rhino.

obucklin (approved):

papachap:
As far as I can see, the current `Feature` implementations do not generate BTLx processings. I assume this is because we haven’t updated them yet to support this workflow?
If that’s the case, do we still want to keep the "Feature" abstraction as a higher-level notation that essentially serves as a way of creating BTLx processings via alternative constructors? 
And if we go in this direction, for which other processings would we want to provide this kind of abstraction? (could/should a tenon/mortise also be represented as a `Feature` that can be added to a `Beam`?)

I am asking this because i am still a bit confused about the relationship between `Feature` and `BTLxProcessings`.

PR #447, opened 2025-06-10
Plate_plate_joint

obucklin:
This adds `PlateJoint` including 2 specific types, `PlateButtJoint` and `PlateMiterJoint`. This seems to be working fairly robustly, but I haven't added a GH workflow yet. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Processings which are not the result of joinery are now serialized with `TimberElement`.
 * Fixed visualization bug in `Plate` due to loft resulting in flipped volume.
 * Fixed a few bugs in the `WallPopulator` workflow including GH component updates.
+* Fixed bug in show_ref_faces GH component.
 
 ### Removed
 
--- a/src/compas_timber/ghpython/components_cpython/CT_Show_Ref_Sides/code.py
+++ b/src/compas_timber/ghpython/components_cpython/CT_Show_Ref_Sides/code.py
@@ -27,7 +27,7 @@ class ShowElementFaces(Grasshopper.Kernel.GH_ScriptInstance):
             ht = 1000
             for side_index in range(len(b.ref_sides)):
                 surface = b.side_as_surface(side_index)
-                ht = min([self.ht, surface.xsize / 6.0, surface.ysize / 6.0])
+                ht = min([ht, surface.xsize / 6.0, surface.ysize / 6.0])
             for side_index in range(len(b.ref_sides)):
                 surface = b.side_as_surface(side_index)
                 frame = b.ref_sides[side_index]


src/compas_timber/connections/solver.py
```diff
@@ -264,3 +272,134 @@ def _exceed_max_distance(pa, pb, max_distance, tol):
...
+        if max_distance is None:
+            max_distance = min(main_plate.thickness, cross_plate.thickness)
+        for pline_a, plane_a in zip(main_plate.outlines, main_plate.planes):
+            for pline_b, plane_b in zip(cross_plate.outlines, cross_plate.planes):
+                line = intersection_plane_plane(plane_a, plane_b, tol=tol)
+                for i, seg_a in enumerate(pline_a.lines):  # TODO: use rtree?
+                    if distance_point_line(seg_a.point_at(0.5), line) <= max_distance:
+                        if is_parallel_line_line(seg_a, line, tol=tol):
+                            if ConnectionSolver.does_segment_intersect_outline(seg_a, pline_b):
+                                return i
```
chenkasirer, line 352: theses seem like potentially a lot of operations at the worst-case scenario. I'd indeed look into RTree or KDTree to see if this can be done in a more efficient manner

perhaps there is some kind of a simplification/assumption we can make to estimate the topology like dunno treating both plates as planes?

src/compas_timber/connections/solver.py
```diff
@@ -264,3 +272,134 @@ def _exceed_max_distance(pa, pb, max_distance, tol):
     @staticmethod
     def _is_near_end(t, length, max_distance, tol):
         return abs(t) * length < max_distance + tol or abs(1.0 - t) * length < max_distance + tol
+
+    def find_plate_plate_topology(self, plate_a, plate_b, tol=TOLERANCE):
```
chenkasirer, line 276: I assume this is the public interface method for the plate intersection topology finding, so I'd mark all the others as private. Also I'd suggest to move all these into a new sovler class that focuses on plates, since this one really only deals with beams.

src/compas_timber/connections/plate_miter_joint.py
```diff
@@ -0,0 +1,49 @@
...
+
+class PlateMiterJoint(PlateJoint):
+    """Creates a mitered edge plate-to-plate connection."""
+
+    def __repr__(self):
+        return "PlateMiterJoint({0}, {1}, {2})".format(self.plate_a, self.plate_b, JointTopology.get_name(self.topology))
+
+    def _adjust_plate_outlines(self):
+        """Adjust the outlines of the plates to match the joint."""
+        print("Adjusting plate outlines for PlateMiterJoint...")
```
chenkasirer, line 15: please remove

src/compas_timber/connections/plate_joint.py
```diff
@@ -0,0 +1,253 @@
...
+from .solver import ConnectionSolver
+
+
+class InterfaceRole(object):
+    """
+    Enumeration of the possible interface roles.
+
+    Attributes
+    ----------
+    MAIN : literal(0)
```
chenkasirer, line 20:
```suggestion
    MAIN : literal("MAIN")
```

src/compas_timber/connections/plate_joint.py
```diff
@@ -0,0 +1,253 @@
...
+        The interface is the main interface.
+    CROSS : literal(1)
```
chenkasirer, line 22:
```suggestion
    CROSS : literal("CROSS")
```

src/compas_timber/connections/plate_joint.py
```diff
@@ -0,0 +1,253 @@
...
+        The interface is the cross interface.
+    NONE : literal(2)
```
chenkasirer, line 24:
```suggestion
    NONE : literal("NONE")
```

src/compas_timber/connections/solver.py
obucklin: This really needs to look at each segment of the polyline that defines the plate to determine which segment to alter for the joint. I was using `is_colinear_line_line`, instead of the combination of `distance_point_line` and `is_parallel_line_line` but it was difficult to control for max distance in a meaningful way. The one thing I could do is only test one of the outlines for each plate, but that can result in missed connections depending on how the plates are offset. This way the solver is a little bit more robust, IMO.

src/compas_timber/utils/__init__.py
```diff
@@ -368,6 +372,59 @@ def correct_polyline_direction(polyline, normal_vector, clockwise=False):
...
+    plane = Plane.from_points(polyline.points)
+    pt = polyline.lines[segment_index].point_at(0.5)
+    perp_vector = Vector(*cross_vectors(polyline.lines[segment_index].direction, plane.normal))
+    point = pt + (perp_vector * 0.1)
+    if is_point_in_polyline(point, polyline):
+        return Vector.from_start_end(point, pt)
+    return Vector.from_start_end(pt, point)
+
+
+def is_point_in_polyline(point, polyline, in_plane=True, tol=TOL):
```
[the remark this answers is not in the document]
obucklin: I could add some checks to make sure the polyline is closed, planar, not self-intersecting... LMK

src/compas_timber/elements/plate.py
```diff
@@ -243,10 +357,18 @@ def shape(self):
             The shape of the element.
 
         """
-        plate_geo = Brep.from_loft([NurbsCurve.from_points(pts, degree=1) for pts in (self.outline_a, self.outline_b)])
+        outline_a = correct_polyline_direction(self.outline_a, self.frame.normal, clockwise=True)
+        outline_b = correct_polyline_direction(self.outline_b, self.frame.normal, clockwise=True)
+        plate_geo = Brep.from_loft([NurbsCurve.from_points(pts, degree=1) for pts in (outline_a, outline_b)])
         plate_geo.cap_planar_holes()
-        if plate_geo.volume < 0:
-            plate_geo.flip()
```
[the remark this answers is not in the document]
obucklin: I've done a fair amount of testing and reconfiguring to make sure that `outline_a`, `outline_b`, `self.normal`, `self.frame` etc are consistent. we could add it as an extra check, but I don't think it's necessary.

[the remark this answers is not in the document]
obucklin: for clarity, I decided to leave all the input curves in their original orientation, and use a reversed copy when necessary to create positive solids in `Plate.shape` and `Plate.geometry`. Previously I was changing the input polylines in the `Plate` constructors, so doing the check here keeps it simple and helps avoid those mixups.

src/compas_timber/elements/plate.py
```diff
@@ -150,6 +167,80 @@ def ref_sides(self):
...
+        ----------
+        feature : :class:`Feature`
+            A feature
+
+        Returns
+        -------
+        None
+
+        """
+        self._features.append(feature)
```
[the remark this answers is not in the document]
obucklin: So this is a bit tricky because we decided to use `Plate._features` for added features and `Plate.features` combines those with the outline (and now cutout) features, which do not get deleted when clearing features. `Element` sets `self.features = []` in its `__init__`, so I had use a setter to have `Element.features` only set `Plate._features`. Let me know if you think there's a better way.

obucklin:
Hey @chenkasirer, thanks for the review. I made some changes as you requested, and I left a few comments of my own. I know this is a bigger PR, so no stress getting it merged. Please take a look when you get a chance!

src/compas_timber/utils/__init__.py
```diff
@@ -368,6 +372,59 @@ def correct_polyline_direction(polyline, normal_vector, clockwise=False):
...
+    Returns
+    -------
+    bool
+        True if the point is inside the polyline, False otherwise.
+    """
+    frame = Frame.from_points(*polyline.points[:3])
+    xform = Transformation.from_frame_to_frame(frame, Frame.worldXY())
+    pgon = Polygon([pt.transformed(xform) for pt in polyline.points[:-1]])
+    pt = point.transformed(xform)
+    if in_plane and not tol.is_close(pt[2], 0.0):
```
chenkasirer, line 423:
```suggestion
    if in_plane and not tol.is_zero(pt[2]):
```

src/compas_timber/utils/__init__.py
```diff
@@ -368,6 +372,59 @@ def correct_polyline_direction(polyline, normal_vector, clockwise=False):
...
+    ----------
+    polyline : :class:`compas.geometry.Polyline`
+        The polyline to check. Must be closed.
+    segment_index : int
+        The index of the segment in the polyline.
+
+    Returns
+    -------
+    int
+        The index of the point in the polyline, or None if not found.
```
chenkasirer, line 389: returns `Vector`, right?

CHANGELOG.md
```diff
@@ -41,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Processings which are not the result of joinery are now serialized with `TimberElement`.
 * Fixed visualization bug in `Plate` due to loft resulting in flipped volume.
 * Fixed a few bugs in the `WallPopulator` workflow including GH component updates.
+* Changed how input `Polyline`s are handled in `Plate`.
```
chenkasirer, line 48: what's the meaning of the change for a user? if public interface changed somehow than perhaps mention it here.

src/compas_timber/utils/__init__.py
obucklin: Copy Paste error or copilot error?

PR #454, opened 2025-06-30
Compas model update

papachap:
This PR is an attempt to align with the current development of compas_model. It removes the old version pin (now set to the latest), refactors affected parts for compatibility, and adapts to recent API changes.

- [X] Breaking change: bug fix or new feature that involve incompatible API changes.

### Checklist

### DIFF
--- a/src/compas_timber/elements/timber.py
+++ b/src/compas_timber/elements/timber.py
@@ -60,6 +60,7 @@ class TimberElement(Element):
         return self._frame
 
     @frame.setter
+    @reset_computed
     def frame(self, frame):
         self._frame = frame
 
@@ -70,6 +71,7 @@ class TimberElement(Element):
         return self._features
 
     @features.setter
+    @reset_computed
     def features(self, features):
         self._features = features
 
@@ -102,6 +104,7 @@ class TimberElement(Element):
         if not isinstance(features, list):
             features = [features]
         self._features.extend(features)  # type: ignore
+        self._geometry = None  # reset geometry cache TODO: should we do that?
 
     @reset_computed
     def remove_features(self, features=None):
@@ -120,3 +123,4 @@ class TimberElement(Element):
             if not isinstance(features, list):
                 features = [features]
             self._features = [f for f in self._features if f not in features]
+        self._geometry = None  # reset geometry cache TODO: should we do that?
--- a/src/compas_timber/elements/timber.py
+++ b/src/compas_timber/elements/timber.py
@@ -23,11 +23,13 @@ class TimberElement(Element):
     @property
     def __data__(self):
         data = super(TimberElement, self).__data__
+        data["frame"] = self.frame
         data["features"] = [f for f in self.features if not f.is_joinery]  # type: ignore
         return data
 
     def __init__(self, features=None, **kwargs):
         super(TimberElement, self).__init__(**kwargs)
+        self._frame = kwargs.get("frame", None)
         self._features = features or []
         self.debug_info = []
 
@@ -52,13 +54,30 @@ class TimberElement(Element):
         return False
 
     @property
+    def frame(self):
+        # type: () -> Frame
+        """The local coordinate frame of the element."""
+        return self._frame
+
+    @frame.setter
+    def frame(self, frame):
+        self._frame = frame
+
+    @property
     def features(self):
         # type: () -> list[Feature]
+        """A list of features applied to the element."""
         return self._features
 
     @features.setter
     def features(self, features):
         self._features = features
+
+    @property
+    def geometry(self):
+        if self._geometry is None:
+            self._geometry = self.compute_elementgeometry()
+        return self._geometry
 
     def remove_blank_extension(self):
         pass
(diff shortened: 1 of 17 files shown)


obucklin (commented):
I don't feel at all qualified to review this. can we get Tom to take a look? what is our strategy here?

src/compas_timber/elements/timber.py
```diff
@@ -51,15 +56,95 @@ def is_group_element(self):
...
+    @frame.setter
+    def frame(self, frame):
+        # type: (Frame) -> None
+        self._frame = frame
+
+    @property
+    def transformation(self):
+        # type: () -> Transformation
+        """The transformation that transforms the element's geometry to the model's coordinate system."""
+        return Transformation.from_frame(self.frame) if self.frame else Transformation()
```
obucklin, line 74: This will definitely not work with our `Plate` because the plate outlines are defined in global coordinates while the Frame is a calculated from those outlines. Looking at the compas_model code, it seems like they only consider using the `element.transform` on the geometry, but we need to use many different properties in global space such as for `ConnectionSolver`. For this to work with plate, I think we need something like a `plate.local_outlines` property to store the geometry in the frame of the plate, and then `plate.outline_a` and `plate.outline_b` would apply the transformation to the local outlines.

obucklin:
I can't get Beam Trasformation to work.
```
>>> frame_a = Frame(Point(100,100,100),Vector(0,0,1), Vector(1,0,0))
>>> frame_b = Frame(Point(1000,1000,100),Vector(1,0,1), Vector(1,1,0))
>>> x= Transformation.from_frame_to_frame(frame_a,frame_b)
>>> beam.transform(x)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\obucklin\Miniconda3\envs\ct-dev\lib\site-packages\compas_model\elements\element.py", line 538, in transform
    self.transformation = transformation * self.transformation
AttributeError: can't set attribute 'transformation'
```

src/compas_timber/elements/timber.py
obucklin: > I agree. the outlines should be saved at coordinates which are with-respect-to the frame which is created from the world space outlines. when accessed from outside beam they should be translated to model space to be useful.
I will implement this in another PR.

> I think it would make sense to expose this as `Plate.from_outlines` which would reference `Beam.from_centerline`. I also think that it should be possible to create a Plate from frame, length, width, height i.e. as a blank. perhaps this should be the default __init__. 
I disagree on this point. I think the beam is fundamentally an extrusion with a cross section and a length. We use width and height because most raw timber comes with a rectangular section. The plate is fundamentally a shape cut out of a thick plane. In both `Plate` and `Beam` cases, the blank is a result of operations we apply to the element. I think an alternate constructor `Plate.from_box(f,l,w,h) would make sense, but not the `__init__`.

src/compas_timber/elements/timber.py
```diff
@@ -51,15 +60,105 @@ def is_group_element(self):
...
+        """The local coordinate system of the element."""
+        return self._frame
+
+    @frame.setter
+    def frame(self, frame):
+        # type: (Frame) -> None
+        self._frame = frame
+
+    @property
+    def transformation(self):
```
obucklin, line 75: as discussed briefly with @papachap I think this needs a setter that transforms the `Beam.frame` to comply with `Element.transform`, `transformed`, etc. methods.

src/compas_timber/elements/timber.py
```diff
@@ -51,15 +60,105 @@ def is_group_element(self):
     def is_fastener(self):
         return False
 
+    @property
+    def frame(self):
+        # type: () -> Frame | None
+        """The local coordinate system of the element."""
+        return self._frame
```
obucklin, line 67: `TimberElement._frame` is never declared and doesn't exist in `Element`. I think we should have this frame functionality as generic as possible, meaning we implement it here and not in `Beam`, `Plate`, etc. I will propose a solution in my PR to this PR.

obucklin: yeah. I implemented this in #535, which just suggests a couple things to this PR

papachap: @obucklin You can review the changes I made based on your feedback and our discussions. Due to the fact that i had already committed the changes that you did in #535 together with some of mine, i haven't merged your PR into this, therefore I will go on and close that PR.

In regards to your last comment about the extensions overriding the `frame` of the beam, I don't think this is happening because extensions are applied dynamically on each property access. The `frame` only gets manipulated when the `transformation` is **set**. In the case of beam_extensions they are only applied when you **get** the transformation via the **getter**, therefore the `frame` remains unchanged.
To proove that, I added a couple of unittests you can have a look.

src/compas_timber/elements/timber.py
obucklin: As discussed with @papachap, in my opinion, we need to align as closely as possible with compas_model as possible, especially in terms of code _interface_. The `Element.frame` returns the Frame in global/model space, whereas we use it to set the local transformation to parent space. this means if we call `element.frame` on elements in a list of a mix of `Element` and `TimberElement` the results will not represent the same thing(assuming we set `TimberElement.frame = frame`. Another issue is if we change the `parent.transformation`, that will not change the model level `element.transformation`/`element.frame`

I see a few options:

1. adopt the `transformation` argument like in `Element.__init__`
2.  make sure it is clear that the `TimberElement` `frame` argument is the local frame, and that the `TimberElement.frame` is not necessarily the same

These are just a few things i noticed while trying to implement a `ContainerElement` class that implements heirarchy in the model tree.

papachap:
@obucklin @chenkasirer The elephant is ready for review again..! Don't get scared by the 7000+ lines of code changes, at least 6000 of those are from the new json and btlx files that we store for testing.

obucklin (approved):
APPROVE!

PR #491, opened 2025-07-14
fix ball_node_joint and add Joint.from_element_list constructor.

obucklin:
This fixes the `fastener.geometry` creation in `BallNodeJoint`. It also introduces a new generic joint constructor `Joint.from _element_list()` which is necessary to have a generic joint constructor that doesn't need a model. `Joint.create()` will call this internally, and the upcoming JointRule workflow rework will use this as well.

### Checklist

### DIFF
--- a/src/compas_timber/connections/joint.py
+++ b/src/compas_timber/connections/joint.py
@@ -189,13 +189,12 @@ class Joint(Interaction):
         model.add_joint(joint)
         return joint
 
-
     @classmethod
     def from_element_list(cls, elements, **kwargs):
         """Creates an instance of this joint from a list of elements. This should be overridden when a joint type takes a list of elements.
 
         This code does not verify that the given elements are adjacent and/or lie in a topology which allows connecting
-        them. This is the responsibility of the calling code.
+        them. This is the responsibility of the calling code. TODO: evaluate whose responsibility this is.
 
         Parameters
         ----------
--- a/src/compas_timber/connections/joint.py
+++ b/src/compas_timber/connections/joint.py
@@ -185,6 +185,31 @@ class Joint(Interaction):
 
         """
 
-        joint = cls(*elements, **kwargs)
+        joint = cls.from_element_list(elements, **kwargs)
         model.add_joint(joint)
         return joint
+
+
+    @classmethod
+    def from_element_list(cls, elements, **kwargs):
+        """Creates an instance of this joint from a list of elements. This should be overridden when a joint type takes a list of elements.
+
+        This code does not verify that the given elements are adjacent and/or lie in a topology which allows connecting
+        them. This is the responsibility of the calling code.
+
+        Parameters
+        ----------
+        *elements : :class:`~compas_model.elements.Element`
+            The elements to be connected by this joint. The number of elements must comply with the `Joint` class's
+            `MIN_ELEMENT_COUNT` and `MAX_ELEMENT_COUNT` attributes.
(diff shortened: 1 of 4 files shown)


src/compas_timber/connections/joint.py
```diff
@@ -185,6 +185,30 @@ def create(cls, model, *elements, **kwargs):
...
+            Additional keyword arguments that are passed to the joint's constructor.
+
+        Returns
+        -------
+        :class:`compas_timber.connections.Joint`
+            The instance of the created joint.
+
+        """
+
+        return cls(*elements, **kwargs)
```
chenkasirer, line 214: it needs to somehow be clear that these joints are not added to the model, i.e. don't joint the beams in the graph.
we didn't properly enforce it before either but ideally there's only one way to create joints and it results in proper connections in the model.

src/compas_timber/elements/fasteners/ball_node_fastener.py
```diff
@@ -153,7 +153,7 @@ def interface_plate(self):
         """Generate a plate from outline_points, thickness, and holes."""
         if not self.base_interface.outline_points:
             return None
-        outline_points = correct_polyline_direction(self.base_interface.outline_points, Vector(0, 0, 1))
+        outline_points = correct_polyline_direction(self.base_interface.outline_points, Vector(0, 0, 1), clockwise=True)
```
chenkasirer, line 156: would you be awesome if you could add a test for this

obucklin: will do

obucklin: So there is a test for this. What (I assume) we currently can't do is test whether the extrusion from this Polyline is inside-out(volume<0). Please correct me if I am wrong.

src/compas_timber/connections/joint.py
obucklin: > it needs to somehow be clear that these joints are not added to the model, i.e. don't joint the beams in the graph. we didn't properly enforce it before either but ideally there's only one way to create joints and it results in proper connections in the model.

more on this point, I am not sure how we should proceed regarding declaring joints in the containers. As I have mentioned a few times, I would like to skip the `JointDefinition` and just declare joints directly in the `BeamPopulator`/`SlabPopulator`. However there is typically no `TimberModel` in that context... I collect all the joints during population and then retrieve them via `SlabPopulator.joints` after adding the `Slab.elements` to the model. Not sure if you see this as a problem...

chenkasirer: yup I understand the issue. I'm also tempted to get rid of `JointDefinition`. we can give creating `Joint` directly it a try.
however just to play devil's advocate and spread further chaos, which we all know is my job around here ;), I'd say that `JointDefinition` represents a valid and known pattern of deferred instantiation that's part of several known design patterns (e.g. factory, command).

There's even a [builtin python functionality](https://docs.python.org/3/library/functools.html#functools.partial) for that that we can use in a vary similar way (with few small modifications to our `create`) and eventually we can even enforce the call to `Joint.create` over direct instantiation.

chenkasirer: finally, if there's one thing here I think we should keep doing is separate the creation of elements and joints (responsibility of the `SlabPopulator`) and the application of those to the model. in other words, I think it's a good thing the `SlabPopulator` doesn't directly modify the model.

Issue #498, opened 2025-07-22
Only one joint currently supported per element pair


obucklin:
or maybe `model.remove_joint` removes all joints between pair of elements.

**current behavior:**
```
generic_joint = GenericJoint([beam_a, beam_b])
joint = cls.create(model, *generic_joint.elements, **kwargs)
model.remove_joint(generic_joint)
```
`model.joints` returns an empty set.

```
generic_joint = GenericJoint([beam_a, beam_b])
model.remove_joint(generic_joint)
joint = cls.create(model, *generic_joint.elements, **kwargs)
```

`model.joints` returns `set(joint)`

**expected behavior:**
```
generic_joint = GenericJoint([beam_a, beam_b])
joint = cls.create(model, *generic_joint.elements, **kwargs)
model.remove_joint(generic_joint)
```
`model.joints` returns  `set(joint)`.

```
generic_joint = GenericJoint([beam_a, beam_b])
joint = cls.create(model, *generic_joint.elements, **kwargs)
```

`model.joints` returns `set(generic_joint, joint)`

Issue #482, opened 2025-07-11
update compas_model


papachap:
update and pin compas_model version to 0.8.0.

**This is a breaking change. This will make compas_timber no longer suppert IronPython.** 

Should come with a new compas_timber release (#478 )
Also closes issues #437 ,

PR #499, opened 2025-07-22
Joint candidates

chenkasirer:
Adding joint candidates in their own attribute so they can co-exist with joints (`interactions`).
Left the promotion logic out of it, done manually (see example script)

* Added `joint_candidates` property to `TimberModel`.
* Added `add_joint_candidate` method to `TimberModel`.
* Added `remove_joint_candidate` method to `TimberModel`.
* `NBeamKDTreeAnalyzer` now uses `model.joint_candidates` instead of filtering `model.joints`.

### Checklist


src/compas_timber/model/model.py
```diff
@@ -314,6 +326,49 @@ def add_joint(self, joint):
...
+    def remove_joint_candidate(self, candidate):
+        # type: (JointCandidate) -> None
+        """Removes this joint candidate from the model.
+
+        Parameters
+        ----------
+        candidate : :class:`~compas_timber.connections.JointCandidate`
+            The joint candidate to remove.
+        """
+        for interaction in candidate.interactions:
```
obucklin, line 364: please take a quick look at issue #498. I don't know what happens under the hood of the Graph, but it looks like we need to support multiple edges for a pair of elements.

obucklin (commented):
I was like,  "where'd the JointCandidate class go... I guess you just merged that one. Do you want to change the name of the `generic_joint.py` module as well?

Do we want the `JointCandidate` to be persistent? in that case we need to support multiple edges per element pair. Not sure if this is currently possible, but see #498 to see what I mean. Is that what the `"interaction"` and `"candidate"` edge attributes are for?

src/compas_timber/model/model.py
```diff
@@ -314,6 +326,49 @@ def add_joint(self, joint):
...
+            element_a, element_b = interaction
+            edge = (element_a.graph_node, element_b.graph_node)
+            if edge in self._graph.edges():
```
obucklin, line 367: I think we need to do the same thing for `TimberModel.remove_joint()`. As I understand it, that method will just remove any edge between the two elements. It should probably filter to just remove `interactions`.

obucklin: @chenkasirer no action on this in a while. did you happen to take a look at this `remove_joint` behavior? let me know when it's ready for another look.

chenkasirer:
@obucklin this is ready for another look.

Basically, `JointCandidate` and `Joint` instances can co-exist on the model. They can be added and removed independently and only when there are no more attributes on an edge, it can get removed via either `remove_joint_candidate` or `remove_joint`.

papachap (approved):
LGTM

The only thing I would have a look, is modifying the `remove_joint()`'s behavior as mentioned by @obucklin.

There are a couple of things that might need a slight adjustment due to the changes in the latest compas_model version. But this would be taken care of in the #482 PR.

obucklin (approved):

PR #509, opened 2025-08-19
propose slab inherits from Plate

obucklin:
As discussed in the last standup, I am using the mermaid documentation to propose a change in the class structure before changing in code. 

I believe `Slab` should inherit from `Plate`. This gives much more formal flexibility and allows compatibility with `PlateJoint`s. 

The logic of a wall, with a baseline and eventually also justificaiton about that line(left, right, center, etc. ) would be downstreamed to the `Wall`. 

The way it is currently written, `Wall` acts like a `beam`. We eventually need to discuss what to do with Slabs and Walls. I would propose that walls have alternate constructors such as `from_baseline()`, `from_box()`, etc. This of coulrse should be coordinated with innosuisse_cadwork, ukraine, etc. 

### Checklist

### DIFF
--- a/docs/contribution/class_diagrams.rst
+++ b/docs/contribution/class_diagrams.rst
@@ -70,41 +70,42 @@ The elements subsystem contains all the core timber elements that can be modeled
          +blank_length : float
          +width : float
          +height : float
+         +blank : Box
          +ref_frame : Frame
          +is_plate : bool = True
-         +compute_geometry()
-         +compute_aabb()
-         +compute_obb()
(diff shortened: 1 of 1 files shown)


obucklin:
As @chenkasirer mentioned, any element can be a "container" by setting it as parent to the contained elements. therefore I am dropping the `ContainerElement` for the time being. I have implemented the `PlateGeometry` class in #542, changed `Slab` to inherit from that class in #543. 

new mermaid looks like this:
image

PR #522, opened 2025-09-03
added pre commit to repo

chenkasirer:
Inspired by @zeitunik to add pre-commit hook to the project, currently just checks linting and formatting using `ruff` . doesn't fix issues, but will block commit if fails.

To use:
1. install new dev requirements using `pip install .[dev]` or `pip install -e .[dev]`
2. install pre-commit hook using `pre-commit install`

Attempts to commit with linting or formatting errors will result in:
image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,5 +7,7 @@ repos:
   hooks:
     # Run the linter.
     - id: ruff-check
+      args: [--fix]
+
     # Run the formatter.
     - id: ruff-format


obucklin:
If I understand correctly, this prevents me from commiting on my own branches if format and lint checks fail. I am personally opposed to this. Often, especially when I'm working on the train, I just need to commit so I can resume later on another computer. it would be annoying if I had to fix a bunch of formatting errors just to do that. ;-D

zeitunik:
@obucklin as Chen already mentioned ruff will fix most issues automatically when there are any. You will only need to recommit in that case. If there are still some issues that can't be solved immediately you can still force commit with the `--no-verify` flag for git.

obucklin (approved):
While I am the wrong person to understand the changes exactly, I have come around to the intent of this PR and wholeheartedly approve.

PR #530, opened 2025-09-12
Transform feature geometries to local coordinates

papachap:
This PR is part of this #454 and was split for ease of review and should be merged first.

What this PR do, is that it converts the computed feature geometries (ie the `cutting_plane` in `JackRafterCut`, or the `lap_volume` in `Lap`) to the element's local coordinates so that it's geometry computation can happen locally when calling `elementgeometry`. 

Subsequently it's geometry in global/model coordinates applies the model transformation chain to that computed geometry when calling `modelgeometry`.

### Checklist

### DIFF
--- a/src/compas_timber/fabrication/double_cut.py
+++ b/src/compas_timber/fabrication/double_cut.py
@@ -338,7 +338,7 @@ class DoubleCut(BTLxProcessing):
                 None, geometry, "Failed to generate cutting planes from parameters and beam: {}".format(str(e))
             )
         # convert to the local coordinates of the beam
-        cutting_planes = [plane.transformed(beam._transformation_to_local()) for plane in cutting_planes]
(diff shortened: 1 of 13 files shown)


obucklin (commented):
These changes make a lot of sense to push the compatibility with the upcoming `TimberModel` changes. 

I think we should consider whether and when/where we want to cache these geometries. IMO we should do this for performance reasons. I could imagine that a `BTLxProcessing`, which is defined locally on the `TimberElement`, could also have a cached local geometry for generating the geometric features. Alternately, we could cache the `TimberElement.geometry` which is presumably supported in the `compas_model.Element` type with the `dirty` attribute. The problem with only caching the  `TimberElement.geometry` is there is that changes to one feature would mean the geometry for all the features on that element would need to be recalculated.

PR #541, opened 2025-10-01
remove beam.transformation

obucklin:
hey @papachap check this out. So it looks like the blank gets used in various BTLxProcessings in global space, so I reverted to global space. maybe at some point we do a local blank or something. 

anyways I removed beam.transformation and updated the ref_frame. I'll try to stop making chnages on this branch.

### Checklist

### DIFF
--- a/src/compas_timber/elements/beam.py
+++ b/src/compas_timber/elements/beam.py
@@ -152,8 +152,9 @@ class Beam(TimberElement):
         The ref_frame is always in model coordinates.
         TODO: This should be upstreamed to TimberElement once all elements are described using a frame.
         """
-
-        return Frame(self.blank.points[1], self.frame.xaxis, self.frame.zaxis)
+        #NOTE: This does not work with self.frame because self.frame is in parent space, and ref_frame needs to be in model space(for now)
+        #NOTE: compas_model.Element.frame is in the global/model space, and would work, but it is not yet implemented in TimberElement.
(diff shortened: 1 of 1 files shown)


papachap:
Hey @obucklin ! we have discussed this in person, but I am also writing my thoughts here so that we have a record of the discussions.

**- `Blank`: in global space** 
I am not opposed to having the `Blank` in local space instead of model space, and I actually don't have any arguments either for or against it. It was done this way after @chenkasirer 's [comment](https://github.com/gramaziokohler/compas_timber/pull/454#discussion_r2351783861)

**- `Transformation`: not accounting for extensions** 
I am also not opposed to having the `Beam.transformation` not having to deal with extensions (that are essentially results of a `Joint`) and instead have the `ref_frame` take care of that instead. This might help to create a distinction between the properties that are prone to modifications caused by joints and the ones who don't.

**- `ref_frame`: in local space**
The `Beam.ref_frame` cannot be in local space, as this will break our entire Joinery system. This is because via the `ref_sides` (therefore `ref_frame`) we extract geometrical information from the interacting beams that is then being used in the  `BTLxProcessing`(s) alt constructors. 

_**Example:** the `cutting_plane` used for the `JackRafterCut` processing applied on the `main_beam` of a `ButtJoint`, is one of the `ref_sides` of the `cross_beam`._ 


For this to work, this plane would need to be transformed from one beam's coordinate space to the other's (relative transformation). This would therefore require a big API change and this is why I have opted with keeping all this information and geometry in model space.

PR #542, opened 2025-10-03
Plate TimberElement alignment

obucklin:
This aligns the `Plate` definition with a BTLx focused `TimberElement`. It unifies several BTLx specific properites like `ref_frame`, `ref_sides`, and `ref_edges`. The Plate is now more parametric, with the outlines being optional arguments that are defined at worldXY aka in relation to the `Plate.frame`. 

Open questions are still, where do we generate geometry, where are the `blank` and `shape` to be found, what else can we upstream to `TimberElement`. Hopefully this change will help with the #454 PR so that we can make all the necessary changes to be compatible with the new local/model space transformations by isolating those changes to the `TimberElement` class. This PR also changes the inheritance of `Fastener`, `Slab`, and `Wall` to inherit from `compas_model.Element`. 

### What type of change is this?

- [ ] Bug fix in a **backwards-compatible** manner.
- [x] New feature in a **backwards-compatible** manner.
- [x] Breaking change: bug fix or new feature that involve incompatible API changes.
- [ ] Other (e.g. doc update, configuration, etc)

### Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [ ] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [ ] I ran all tests on my computer and it's all green (i.e. `invoke test`).
- [ ] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [ ] I added new functions/classes and made them available on a second-level import, e.g. `compas_timber.datastructures.Beam`.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added necessary documentation (if appropriate)

PR #543, opened 2025-10-05
Slab inherits plate geometry

obucklin:
This is a follow-up to #542 

The `Slab` class inherits from `PlateGeometry` and `compas_model.Element` but not from `TimberElement`. This helps distinguish between "BTLx-able" Elements and other elements. For the moment, I didn't see the need to add an explicit `ContainerElement`, as I think the `SlabPopulators` can take care of implementing the parent/child hierarchy in the model. Where this may change is if we decide to implement a local/sub-model functionality as part of the parent element.

### Checklist

### DIFF
--- a/tests/compas_timber/test_nesting.py
+++ b/tests/compas_timber/test_nesting.py
@@ -421,7 +421,7 @@ def test_nest_per_group_basic():
     model = TimberModel()
     polyline = Polyline([(0, 0, 0), (1000, 0, 0), (1000, 1000, 0), (0, 1000, 0), (0, 0, 0)])
     # Create Group A with 2 beams
-    group_a = Slab(outline=polyline, thickness=0)
+    group_a = Slab.from_outline_thickness(outline=polyline, thickness=0)
     beam_a1 = Beam(frame=Frame.worldXY(), length=2000, width=120, height=60)
     beam_a2 = Beam(frame=Frame.worldXY(), length=1500, width=120, height=60)
     model.add_element(group_a)
@@ -429,7 +429,7 @@ def test_nest_per_group_basic():
     model.add_element(beam_a2, parent=group_a)
 
     # Create Group B with 2 beams
-    group_b = Slab(outline=polyline, thickness=0)
+    group_b = Slab.from_outline_thickness(outline=polyline, thickness=0)
     beam_b1 = Beam(frame=Frame.worldXY(), length=800, width=120, height=60)
     beam_b2 = Beam(frame=Frame.worldXY(), length=1200, width=120, height=60)
     model.add_element(group_b)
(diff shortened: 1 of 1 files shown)


obucklin:
@papachap @chenkasirer This is ready for review. dont forget to look at #542 first

obucklin:
> @obucklin does this need to be rebase on main now that #542 is merged?

@chenkasirer yep I just merged the changes from main. This bring up the issue that `WallPopulator` was not updated with the `compas_model_update` pr and I believe at the very least the beam locations will be wrong due to the `element.transformation` changes. I'm don't really want to spend time patching that when SlabPopulator is being actively reworked, but I'd like to know what you think.

src/compas_timber/elements/slab.py
```diff
@@ -1,284 +1,215 @@
...
+        data["name"] = self.name
         return data
 
-    def __init__(self, outline, thickness, openings=None, frame=None, name=None, **kwargs):
-        # type: (compas.geometry.Polyline, float, list[compas.geometry.Polyline], Frame, dict) -> None
-        super(Slab, self).__init__(name=name)
-        self.outline = outline
-        self.thickness = thickness
-        self.openings = openings or []
+    def __init__(self, transformation, length, width, thickness, local_outline_a=None, local_outline_b=None, openings=None, name=None, **kwargs):
```
papachap: I think we should aim for consistency across the `compas_timber` element classes. If the decision is to maintain a Frame-based API for `TimberElement` (i.e., accepting a `Frame` and converting it to a `Transformation` internally), we should apply the same approach here.

src/compas_timber/elements/slab.py
```diff
@@ -1,284 +1,215 @@
...
+    WALL = "wall"
+    FLOOR = "floor"
+    ROOF = "roof"
+    GENERIC = "generic"
+
+
+class Slab(PlateGeometry, Element):
+    """Represents a timber slab element (wall, floor, roof, etc.).
+
+    Serves as container for beams, joints and other related elements and groups them together to form a slab.
```
chenkasirer: technically, the slab knows nothing about the joints. these are still only described at the model level. the only sub-elements of a slab are other `TimberElement` or `Element` objects.

obucklin:
OK i really don't get why my tests all pass on my machine but not on github...

papachap:
> OK i really don't get why my tests all pass on my machine but not on github...
> 
> especially when the failing tests are related to slot and double cut. I didn't touch either of those modules or their respective tests

it's a breaking change on the latest compas version (2.15.0) that broke our BTLx Processing calculations. Probably in your conda env you have an earlier version of compas. Check @chenkasirer 's message on the slack compas_timber channel.

obucklin:
@chenkasirer @papachap please take a look at this when you get a chance

src/compas_timber/ghpython/components_cpython/CT_Model/code.py
```diff
@@ -58,9 +57,6 @@ def RunScript(
         ##### Adding elements #####
         self.add_elements_to_model(Model, Elements, Containers)
```
papachap:
```suggestion
        self.add_elements_to_model(Model, Elements)
```

I suppose since `containers` input is gone, this should be updated as well.

papachap (approved):
LGTM. I left a small comment in relation to the CTModel gh component, once this is address this is good to go from my side.

PR #544, opened 2025-10-06
Align PlateJoint with Joint and add SlabJoint

obucklin:
This cleans up the `PlateJoint` class and aligns it more with the BTLx focused approach of the `Joint` class. For example, I removed the `PlateToPlateInterface`, which was not being used, and instead, new implementations of plate joints can just apply `BTLxProcessings` as features.

The other side is to make the `SlabConnectionInterface` which is more explicitly a slab-oriented joinery "feature". It doesn't create `BTLxProcessings` directly, but is rather the "feature" that the `SlabJoint` applies to each `Slab` element during processing. This is accomplished via a few methods like `add_extensions`, `add_features`, and `remove_interfaces`. 

### Checklist


src/compas_timber/slab_features/slab_connection_interface.py
```diff
@@ -0,0 +1,70 @@
...
+        data["edge_index"] = self.edge_index
+        data["interface_role"] = self.interface_role
+        return data
+
+    @property
+    def polyline(self):
+        return self._polyline.transformed(self.transformation)
+
+    @property
+    def geometry(self):
```
chenkasirer: should this be added to the interface of `SlabFeature`?

src/compas_timber/connections/slab_joint.py
```diff
@@ -0,0 +1,135 @@
...
+            self.a_segment_index,
+        )
+
+        b_interface_polyline = Polyline(
+            [
+                self.a_outlines[1][self.a_segment_index],
+                self.a_outlines[1][self.a_segment_index + 1],
+                self.a_outlines[0][self.a_segment_index + 1],
+                self.a_outlines[0][self.a_segment_index],
+                self.a_outlines[1][self.a_segment_index],
```
chenkasirer: same here

PR #619, opened 2025-11-20
KTrussButtJoint

ericgozzi:
Joint of K topology, the two main beams are bot chutted with a `DoubleCut` feature whille the cross beam with a `Pocket` feature. The common cutting plane for the two main beams is build along the bisector line between the two main beams.

### Checklist

k-truss-buut

[GH.zip](https://github.com/user-attachments/files/23820895/GH.zip)

### DIFF
--- a/src/compas_timber/connections/butt_joint.py
+++ b/src/compas_timber/connections/butt_joint.py
@@ -15,6 +15,7 @@ from compas_timber.utils import polyhedron_from_box_planes
 from .joint import Joint
 from .solver import JointTopology
 from .utilities import beam_ref_side_incidence
+from .utilities import extend_beam_to_plane
 
 if TYPE_CHECKING:
     from compas_timber.elements.beam import Beam
@@ -160,13 +161,7 @@ class ButtJoint(Joint):
         assert self.main_beam and self.cross_beam
         # extend the main beam
         try:
-            start_main, end_main = self.main_beam.extension_to_plane(self.butt_plane)
-            extension_tolerance = 0.01  # TODO: this should be proportional to the unit used
-            self.main_beam.add_blank_extension(
-                start_main + extension_tolerance,
-                end_main + extension_tolerance,
-                self.guid,
-            )
+            extend_beam_to_plane(beam=self.main_beam, plane=self.butt_plane, joint=self)
         except AttributeError as ae:
             raise BeamJoiningError(beams=self.elements, joint=self, debug_info=str(ae), debug_geometries=[self.butt_plane])
         except Exception as ex:
@@ -175,15 +170,9 @@ class ButtJoint(Joint):
         # extend the cross beam
         if self.modify_cross:
             try:
-                start_cross, end_cross = self.cross_beam.extension_to_plane(self.back_plane)
+                extend_beam_to_plane(self.cross_beam, self.back_plane, self)
             except AttributeError as ae:
                 raise BeamJoiningError(beams=self.elements, joint=self, debug_info=str(ae), debug_geometries=[self.back_plane])
-            extension_tolerance = 0.01  # TODO: this should be proportional to the unit used
-            self.cross_beam.add_blank_extension(
-                start_cross + extension_tolerance,
-                end_cross + extension_tolerance,
-                self.guid,
-            )
 
     def add_features(self) -> None:
         """Removes this joint's previously generated features and adds new features to each beam."""
--- a/src/compas_timber/connections/joint.py
+++ b/src/compas_timber/connections/joint.py
@@ -262,11 +262,11 @@ class Joint(Data):
             The direction of the beam towards the joint.
 
         """
-        end, _ = beam.endpoint_closest_to_point(self.location)
-        if end == "start":
-            beam_direction = beam.centerline.vector
-        else:
+        is_start, _ = beam.endpoint_closest_to_point(self.location)
+        if is_start == "start":
             beam_direction = beam.centerline.vector * -1
+        else:
+            beam_direction = beam.centerline.vector
         beam_direction.unitize()
         return beam_direction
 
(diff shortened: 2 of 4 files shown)


src/compas_timber/connections/k_truss_butt.py
```diff
@@ -0,0 +1,366 @@
...
+            dir2 = beam_2.centerline.vector
+        else:
+            dir2 = beam_2.centerline.vector * -1
+
+        # The bisector direction is the normalized sum of both directions
+        bisector_direction = (dir1 + dir2).unitized()
+
+        # Create rotation plane of the bisector
+        rotation_plane = Plane.from_point_and_two_vectors(intersection_point, dir1, dir2)
+
```
obucklin: this is probably faster if you take cross vectors of the two beams and get cross vector of that with the bisector direction.

src/compas_timber/connections/k_butt.py
```diff
@@ -0,0 +1,412 @@
...
+        """
+
+        if cls.is_cross_beam(beam1, beam2, beam3):
+            return beam1, [beam2, beam3]
+        elif cls.is_cross_beam(beam2, beam1, beam3):
+            return beam2, [beam1, beam3]
+        elif cls.is_cross_beam(beam3, beam1, beam2):
+            return beam3, [beam1, beam2]
+        else:
+            raise ValueError("Could not identify cross beam in K configuration")
```
ericgozzi: @obucklin and @papachap  Would it make sense to have a gneric/abstract parent class `KJoint` that contains these two methods and similars? These two methods could be needed for a lot of other Joints in this topology. 
In this commit I am calling this two methods also from the `KMiterJoint`.

ericgozzi: Thank's a lot @obucklin !
With this method `Cluster.parse_main_and_cross_beams()`  and a override of `promote_cluster` in each KJoint it works perfectly also with the rules.

src/compas_timber/connections/k_miter.py
```diff
@@ -0,0 +1,342 @@
...
+
+        Returns
+        -------
+        :class:`~compas_timber.connections.KMiterJoint`
+            The created joint instance.
+        """
+        if len(cluster.joints) <= 2 or len(cluster.joints) >= 4:  # not a K_TOPO
+            raise BeamJoiningError("Cluster does not represent a K topology joint.")
+        main_beams, cross_beams = cluster.parse_main_and_cross_beams()
+        elements = list(cross_beams) + list(main_beams)
```
obucklin: maybe check that there is only one cross beam.

ericgozzi:
Hello!

Following the concerns about creating _“throw-away”_ joints and the suggestion to expose the relevant methods externally, I came up with the following proposal.
At the moment, this is implemented only for the ButtJoint class.

Instead of instantiating a full joint object, the idea is to directly call the meaningful operation, for example:

```
feature = ButtJoint.cut_main_beam(
    self.cross_beam,
    self.main_beam,
    mill_depth=self.mill_depth
)
self.features.append(feature)
```

This way, the joint class acts also as _"feature generator"_ and must not be fully instantiated.


@obucklin , @chenkasirer , @papachap Before moving forward with this approach, I would really appreciate hearing your thoughts.

ericgozzi:
> @ericgozzi hey whats the status of this. It would be good if you could work on addressing all the comments and then we can get this merged.

@obucklin  It is ready for a new round of review 😊

src/compas_timber/connections/butt_joint.py
```diff
@@ -219,35 +210,80 @@ def add_features(self) -> None:
...
+            _, lap_width = main_beam.get_dimensions_relative_to_side(main_beam_ref_side_index)
+            ref_side_dict = beam_ref_side_incidence(main_beam, cross_beam, ignore_ends=True)
+            cross_beam_ref_side_index = min(ref_side_dict, key=ref_side_dict.get)
+
+            lap_feature = Lap.from_plane_and_beam(cutting_plane, cross_beam, lap_width, mill_depth, ref_side_index=cross_beam_ref_side_index)
+
+            return lap_feature
+
+    @staticmethod
+    def pocket_on_cross_beam(cross_beam: Beam, main_beam: Beam, mill_depth: Optional[float] = None, butt_plane: Optional[Plane] = None, conical_tool: bool = False) -> Pocket:
```
obucklin: Make this a bit clearer. something like `get_cross_beam_pocket_feature`

src/compas_timber/connections/butt_joint.py
```diff
@@ -219,35 +210,80 @@ def add_features(self) -> None:
...
+        self.cross_beam.add_features(pocket)
+        self.features.append(pocket)
+
+    def restore_beams_from_keys(self, model: TimberModel):
+        """After de-serialization, restores references to the main and cross beams saved in the model."""
+        self.main_beam = model[self.main_beam_guid]
+        self.cross_beam = model[self.cross_beam_guid]
+
+    @staticmethod
+    def cut_main_beam(cross_beam: Beam, main_beam: Beam, mill_depth: Optional[float] = None, butt_plane: Optional[Plane] = None) -> JackRafterCutProxy:
```
obucklin: Make this a bit clearer. something like `get_main_beam_jack_cut_feature`

obucklin (changes requested):
Hey @ericgozzi. I looked at this and made some comments. I think that your approach of changing to static methods is great, but also consider if things can be made into functions, too. I think this is part of a much larger refactoring, but it's good to get a start here as a test case. 

Please also add your new joints to the class_diagrams.rst doc.

src/compas_timber/connections/joint.py
```diff
@@ -8,6 +11,11 @@
 
 from .solver import JointTopology
 
+if TYPE_CHECKING:
+    from compas.geometry import Vector
+
+    from compas_timber.elements.beam import Beam
```
papachap: please use second level imports (ie. `from compas_timber.elements import Beam`) as this is the import style followed in the compas ecosystem

PR #646, opened 2025-12-18
reduced the build matrix

chenkasirer:
reduced the size of the CI build matrix to speed up the PR checks and use less resources.

Choose these versions because: `3.9` cause of Rhino, `3.12`  + `3.14` to each represent their release life-cycle phase.

Also removed the Ubuntu runner although maybe not so necessary.

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,8 @@ dev = [
 ]
 
 gh = [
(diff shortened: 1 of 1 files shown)


gonzalocasas (approved):
LGTM!

PR #651, opened 2026-01-08
Some changes for #544

chenkasirer:
some fixes including:
* avoid circular imports
* use `Optional` instead of `Union` where appropriate
* cleaner implementation of __data__ with multiple inheritance.

### DIFF
--- a/src/compas_timber/elements/plate_geometry.py
+++ b/src/compas_timber/elements/plate_geometry.py
@@ -61,7 +61,9 @@ class PlateGeometry(Data):
     @property
     def __data__(self):
         data = super().__data__
-        data.update({"local_outline_a": self._original_outlines[0], "local_outline_b": self._original_outlines[1], "openings": self.openings})
+        data["local_outline_a"] = self._original_outlines[0]
+        data["local_outline_b"] = self._original_outlines[1]
+        data["openings"] = self.openings
         return data
 
     def __init__(self, local_outline_a, local_outline_b, openings=None, **kwargs):


src/compas_timber/fabrication/slot.py
```diff
@@ -16,7 +19,8 @@
 from compas.geometry import intersection_segment_plane
 from compas.tolerance import TOL
 
-from compas_timber.elements.beam import Beam
+if typing.TYPE_CHECKING:
```
obucklin: is this because you hit circular imports or just to avoid importing extra stuff if we aren't type-checking? I assume without the try-except it will still fail if there are circular imports. Do we need to do anything with `__future__.annotations` to allow typing and avoid circular imports?

src/compas_timber/fabrication/free_contour.py
```diff
@@ -1,12 +1,9 @@
-try:
-    from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING
```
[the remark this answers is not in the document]
chenkasirer: right you are. we'd need to add this import in order to have proper python3 type hint.
I didn't bother here though, just removed the `try..except` block which had to do with python2.7 backwards compatibility and less to do with circular imports.

src/compas_timber/fabrication/slot.py
obucklin: @chenkasirer should I go ahead and merge this?

chenkasirer: merged!

Issue #662, opened 2026-01-13
Some tests depend on global Tolerance state and fail when run in isolation


razec3:
**Description**
Some tests modify the value of `compas.tolerance.Tolerance`. Subsequent tests then silently depend on this global state and fail when run in isolation or as subsets of the test suite.

Running the full test suite passes, but selecting specific test files ([test_slot.py](https://github.com/gramaziokohler/compas_timber/blob/680cbdcfcc5bd9a9738da23574aae56997df1d3a/tests/compas_timber/test_slot.py) and [test_utils.py](https://github.com/gramaziokohler/compas_timber/blob/680cbdcfcc5bd9a9738da23574aae56997df1d3a/tests/compas_timber/test_utils.py)) causes 4 tests to fail, indicating order-dependent behavior.

Affected tests:
- test_slot.py
  - test_slot_apply_points
  - test_slot_apply_frames
  - test_slot_volume_subtracting_polyhedron
- test_utils.py
  - test_intersection_line_plane_param

**To Reproduce**
1. Run full test suite: `invoke test` or `pytest tests` --> all tests pass
2. Run subset of tests, e.g. `pytest tests/compas_timber/test_slot.py` --> multiple tests fail
3. Running `pytest tests/compas_timber/test_utils.py` shows similar behavior

**Expected behavior**
If all tests pass when running the full test suite, they should also pass when run individually or as subsets, independent of execution order.

**Desktop (please complete the following information):**
- OS: Windows
- Python version: 3.10.19
- Python package manager: conda

**Additional context**

While debugging, I inspected the value of `compas.tolerance.Tolerance` immediately before a failing test.

When running the full test suite:
`
    Tolerance(
        unit='M',
        absolute=0.001,
        relative=0.01,
        angular=1e-06,
        approximation=0.001,
        precision=3,
        lineardeflection=0.001,
        angulardeflection=0.1
    )
`

When running the same test in isolation:
`
    Tolerance(
        unit='M',
        absolute=1e-09,
        relative=1e-06,
        angular=1e-06,
        approximation=0.001,
        precision=3,
        lineardeflection=0.001,
        angulardeflection=0.1
    )
`

This difference explains the numerical assertion failures and confirms that global tolerance state is not reset consistently between tests.

obucklin:
@chenkasirer @papachap this was solved with changes to compas.tolerance, no?

chenkasirer:
[...]

PR #663, opened 2026-01-13
Test/662 reset tolerance state

razec3:
### Description
This PR follows up on #662, and fixes order-dependent test behavior caused by shared global state in `compas.tolerance.Tolerance`.
The goal is to make tolerance assumptions explicit and restore proper test isolation, without changing the underlying behavior being tested.

A global pytest fixture was added to reset the tolerance before each test, ensuring a clean and consistent baseline. (tests/compas_timber/conftest.py) 
Tests that previously relied on a modified tolerance are updated to set the required tolerance explicitly, using the same values they were implicitly relying on before. (previous implicit value was obtained through debugging)

Now tests pass both as a full suite and when run in isolation.

No production code is affected.

### Checklist

### DIFF
--- a/tests/compas_timber/conftest.py
+++ b/tests/compas_timber/conftest.py
@@ -1,0 +1,13 @@
+from compas.tolerance import TOL
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def reset_tolerance():
+    """
+    Reset the global tolerance before each test.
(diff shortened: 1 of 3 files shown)


chenkasirer (approved):
@razec3 thanks for this, LGTM!

Just left a couple tiny annoying comments.

PR #667, opened 2026-01-19
sorted out attributes situation across the different timber elements

chenkasirer:
This PR sorts out the discrepancies related to the `attributes` dictionary which can be useful for project specific attributes and was only available on some element types and was sometimes improperly serialized.

### Checklist

### DIFF
--- a/tests/compas_timber/test_panel.py
+++ b/tests/compas_timber/test_panel.py
@@ -128,7 +128,8 @@ def test_panel_serialization_with_attributes():
     assert deserialized.attributes["custom_attribute"] == "custom_value"
     assert deserialized.attributes["another_attribute"] == 42
 
-    def test_from_outline_thickness():
+
+def test_from_outline_thickness():
     polyline_a = Polyline([Point(0, 0, 0), Point(0, 20, 0), Point(10, 20, 0), Point(10, 0, 0), Point(0, 0, 0)])
     polyline_b = Polyline([Point(0, 0, 1), Point(0, 20, 1), Point(10, 20, 1), Point(10, 0, 1), Point(0, 0, 1)])
     panel_a = Panel.from_outlines(polyline_a, polyline_b)


src/compas_timber/elements/timber.py
```diff
@@ -71,10 +71,13 @@ def __data__(self):
         data["width"] = self.width
         data["height"] = self.height
         data["features"] = [f for f in self.features if not f.is_joinery]  # type: ignore
+        data.update(self.attributes)
         return data
 
     def __init__(self, frame, length, width, height, **kwargs):
         super().__init__(transformation=Transformation.from_frame(frame), **kwargs)
+        self.attributes = {}
+        self.attributes.update(kwargs)
```
obucklin: So as I understand it, `self.attributes` will just catch and store all the kwargs. Will this lead to values being stored in two places as TimberElement attributes and in the `TimberElement.attributes` dict? Is there a chance that `self.attributes` and `**kwargs` get mixed up when serializing and lead to attributes getting passed as kwargs? or is that the desired behavior?

Similarly, does this mean I can use an item typically found in the `attributes` dict and use it as a kwarg in the TimberElement constructor, eg `beam=Beam.from_endpoints(point_a,point_b, width = 1.0, height=2.0, category="my_category")` instead of creating the beam first and then changing the category by explicitly calling `beam.attributes["category"]` (or `get("category")`)?

obucklin (approved):
looks good to me, although I do have a question for my own comprehension below.

src/compas_timber/elements/timber.py
chenkasirer: indeed, attributes that are part of `Element.__data__` and are not consumed by `Beam` or `TimberElement` will end up in `attributes` that way, essentially creating a duplication. However, since we're taking a rather selective approach to using `Element` I took the liberty of removing its `__data__` from the chain. this rids us from this happening and also removes some hacky boilerplate from the data interface implemetnation in `TimberElement`.

PR #682, opened 2026-02-19
Added plate from brep

### LINKED ISSUE #458: plate from brep
generate a plate element from a brep. needs a fairly specific geometry to work. needs 2 parallel faces with same number of edges

nbbosco:
closes https://github.com/gramaziokohler/compas_timber/issues/458

This pull request introduces how plate and panel elements are constructed from breps in the codebase. It adds robust support for multi-face breps, refactors constructors for clarity and parallelism between plates and panels, and centralizes brep parsing logic in a new utility. There are also enhancements to utility functions for polyline extraction and validation, as well as improved handling of degenerate geometry.

**Plate and Panel Construction Improvements**
* Added new multi-face brep support via the `Plate.from_brep()` and `Panel.from_brep()` class methods, which automatically create elements by detecting parallel faces. New `from_face_thickness()` methods replace the old single-face brep constructors for both plates and panels. 
* Breaking change: `Plate.from_brep()` now exclusively constructs plates from multi-face breps, delegating brep parsing to the new utility. Existing code using single-face breps must switch to `from_face_thickness()` methods.

**Utility Function Enhancements**
* Added `get_plate_geometry_outlines_from_brep` to `compas_timber.utils`, a shared utility for extracting outlines and openings from multi-face breps using mesh-based topology. 
* Refactored `polyline_from_brep_loop` and `polylines_from_brep_face` to use improved joining logic, filter degenerate segments, and handle curved edges by requiring pre-tessellation.
* Improved `join_polyline_segments` to silently discard degenerate (zero-length) segments before joining.

**Validation and Robustness**
* Strengthened outline validation in `_check_outlines` to ensure outlines are planar, parallel, and correctly positioned on the XY-plane.
* Improved point-in-polyline logic to handle degenerate polygons and ensure proper transformation and normal calculation.

**API Consistency and Documentation**
* Updated docstrings and method signatures for clarity and consistency between plates and panels, reflecting the new construction logic and parallel API. 

These changes collectively provide a more robust, flexible, and maintainable approach to brep-based element construction and geometry extraction.

### Checklist

image


src/compas_timber/elements/plate_geometry.py
```diff
@@ -256,7 +256,9 @@ def _check_outlines(outline_a: Polyline, outline_b: Polyline) -> None:
             raise ValueError("The outline_b is not closed.")
         if len(outline_a) != len(outline_b):
             raise ValueError("The outlines must have the same number of points.")
-        if all(not TOL.is_close(p[2], 0) for p in outline_a.points):
-            raise ValueError("outline_a must be planar. Polyline: {}".format(outline_a))
-        if all(not TOL.is_close(p[2], outline_b[0][2]) for p in outline_b.points):
+        # Check if outline_a is planar (all z-coordinates are the same)
+        if not all(TOL.is_close(p[2], outline_a[0][2]) for p in outline_a.points):
```
obucklin: Per the definition of the `PlateGeometry`, `outline_a` must be on the XY-plane, so the Z-values must be 0.

src/compas_timber/utils/__init__.py
```diff
@@ -371,7 +371,24 @@ def is_point_in_polyline(point, polyline, in_plane=True, tol=TOL):
...
+    if b is None:
+        return False
+    for pt in pts:
+        cross = cross_vectors(subtract_vectors(b, a), subtract_vectors(pt, a))
+        if length_vector(cross) > TOL.absolute:
+            c = pt
+            break
+    if c is None:
+        return False
+    frame = Frame.from_points(a, b, c)
```
obucklin: So if I understand correctly, lines 374-391 make sure that the points used to make the `frame` are not coincident or colinear. please add comments to make it clear what you are checking for.

src/compas_timber/utils/__init__.py
```diff
@@ -371,7 +371,24 @@ def is_point_in_polyline(point, polyline, in_plane=True, tol=TOL):
     bool
         True if the point is inside the polyline, False otherwise.
     """
-    frame = Frame.from_points(*polyline.points[:3])
+    pts = list(polyline.points)
+    a = pts[0]
+    b = None
+    c = None
+    for pt in pts[1:]:
+        if distance_point_point(a, pt) > TOL.absolute:
```
obucklin: `distance_point_point` has a square root, which is fairly heavy. consider using `distance_point_point_squared` or just subtracting the vectors and checking if all values are 0. you can do this with `if not TOL.is_allclose(pt, a):` which just compares the values of the two points directly, which is probably the fastest.

src/compas_timber/utils/__init__.py
```diff
@@ -591,19 +608,52 @@ def add_seg_to_points(seg: Line, points: list) -> bool:
...
+        cannot be joined into a single closed polyline.
     """
-    segments = [Line(edge.start_vertex.point, edge.end_vertex.point) for edge in loop.edges]
-    return join_polyline_segments(segments, close_loop=True)
+    def _edge_points(edge):
+        start = edge.start_vertex.point
+        end = edge.end_vertex.point
+        is_degenerate = TOL.is_allclose(start, end)
+        try:
+            pts = list(edge.curve.to_polyline(num_curve_samples).points)
```
obucklin: This uses the Rhino function and will not be testable in pure python script. Curved brep edges tend to imply that we have curved surfaces, which we can't really handle(with the exception of a ruled surface between 2 lines as a `DualContour`). If you have a curved curve that gets lofted to a straight one as one of your BrepFaces, then you will have a point mismatch between the two face outlines. I would suggest that we make any use of curved curves have to take place outside of compas_timber functions, so that we only take polygon/polyline BrepFace geometries in this method.

obucklin (changes requested):
Hey @nbbosco, nice work on this. I left a lot of comments, and I think this needs some careful consideration of what we can support. Here are my general thoughts:

- This should be a constructor for both `Plate` and `Panel`. Essentially the function could be `get_plate_geometry_outlines_from_brep` or something similar.
- The `PlateGeometry` is highly constrained by what we can produce/what the BTLx FreeContour can handle, and our implementation of it. I think you could do a lot with the Data structure of the Brep. essentially you have the top and bottom surface, whose outlines make up the inputs for Platte and Panel. Every other BrepFace can only have 4 edges and has to link the top and bottom surface. You can use that knowledge to parse the faces with very little geometric analysis. I don't know if there is a `Brep.to_mesh` that would cleanly translate the faces to a mesh data structure, but it could potentially be useful.

obucklin (changes requested):
Hey @nbbosco I left some comments. I think you can still do more with the data before looking at the geometry. In `get_plate_geometry_outlines_from_brep` it would help a lot if you could add more comments and/or name your variables better, so that it is easier to understand what you are doing. 

I think the `mesh_from_brep_simple` might get upstreamed to compas.geometry, but we will see. maybe we change that in the next PR. for now keep using this version.

keep it up, almost there.

obucklin (approved):
Hey nice work! LGTM just make sure you update any GH components in timber_design repo

PR #737, opened 2026-04-27
Joint local planes

obucklin:
This fixes a problem with user input `Plane` arguments in `LMiterJoint`, `TButtJoint` and `LButtJoint`. Previously these planes were only defined in global space, and when elements were transformed or one model was combined with another, these planes would not be transformed accordingly. 

My solution is as follows. Each of the affected joints gets a `create()` override which takes care of transforming the plane to local joint space, which I define with the `beam_a` local space. Each `__init__()` method of the affected joints takes a `local_*_plane`, which is the core parametric definition of the plane in the joint. This will require a change to the GH components which will have to check the `create` overrides to get the correct arguments rather than the `__init__`.

I have thought about a couple other options to solve this:
**Option B**
We pass the global plane to the `__init__` and then transform to local space inside the `__init__`. Both local and global planes would be stored and serialized. These would be passed to the `__init__` on deserialization, and the `__init__` would parse the values for generating the local data. This is somewhat complex, because the local frame depends on the joint elements, but those don't exist when `__init__` is called during deserialization. This also creates some confusing arguments for the `__init__` method which would also end up in the GH component params.

**Option C**
We pass the global plane to the `__init__` and then transform to local space inside the `__init__`. Only the local plane would be stored and serialized. The joint would then get a `from_data` override which would use the local plane, while normal instantiation would use the global plane. I will implement this so we can compare. 

 

### Checklist


chenkasirer:
@obucklin @papachap  So if I had to choose one of the 3 options, I think I would go with option A (model to local in `Joint.create()`).
However, I'd like to push back a bit and point out that I think passing an arbitrary butt plane is fundamentally problematic and is the root cause of this pickle.

I think a better solution, one that's more consistent with the **parametric** way we usually define our joint in is to always use some standard plane, in the local space of one of the beams, and parametrically define how we'd like it to be modified. for example, we always start with one of the ref_sides and allow passing one or two rotation angles (something akin to BTLx's `angle` and `inclination`). these two angles can also be calculated from an arbitrary - model space - plane, prior to creating the joint, allowing the same design freedom.

Then these become simple data values that are easy to serialize and use, don't require any trickery, and are easier to validate agains erroneous edge cases. what do you think? 

Another thought that came up in discussion with @papachap: if the arbitrary plane is the result of, for example, a cut needed due to there being a wall there, then it's arguable whether it should be placed there by the joint - rather it should be made a `JackRafterCut` and added directly to the respective beams.

src/compas_timber/connections/l_butt.py
```diff
@@ -95,13 +113,69 @@ def main_beam_ref_side_index(self):
...
+        angle, offset = decompose_plane_to_ref_side(ref_side, back_plane, plane_name="back_plane", reference_name="main_beam")
+        joint.back_plane_ref_side_index = ref_side_index
+        joint.back_plane_angle = angle
+        joint.back_plane_offset = offset
+
     @classmethod
     def create(
         cls, model, main_beam=None, cross_beam=None, mill_depth=None, small_beam_butts=False, modify_cross=True, reject_i=False, butt_plane=None, back_plane=None, **kwargs
     ):
+        """Creates an instance of this joint and adds it to the model.
```
chenkasirer: I don't think `create` should define its own arguments that diverge from the Joint's `__init__`, that would be imho a major shift in our API design. 
`create`'s responsibility is a clear one: making sure the joint is created in a consistent way, i.e. added to the model. arguments are passed through.
I think therefore we should override it very sparingly. in this case the plane should get deconstructed externally (e.g. in the GH component)

chenkasirer (changes requested):

src/compas_timber/connections/l_butt.py
obucklin: We do already override `create()` for `BallNodeJoint` and `LButtJoint` because we then parse/preprocess arguments before passing to `__init__()`

obucklin:
@chenkasirer ok I implemented without overriding `create`. ready for another review when you have time.

src/compas_timber/connections/l_butt.py
chenkasirer: my reservations are about mismatch of the arguments expected in `create` vs. `__init__`. in `BallNodeJoint` and `LButtJoint` this is not the case.

obucklin:
@chenkasirer I went ahead and merged your suggestions. let me know if there's anything else.

chenkasirer (approved):
LGTM

PR #740, opened 2026-04-28
Remove openings from PlateGeometry

obucklin:
This is part of the push for the PanelPopulator.

Openings (windows, doors) were previously stored as raw polylines on PlateGeometry and applied directly during geometry computation. This PR moves that responsibility to the feature system, where it belongs.

**What changed:**

- PlateGeometry, Plate.__init__, and Panel.__init__ no longer accept an openings parameter. The openings attribute is gone from PlateGeometry.
- Plate.from_outlines(openings=[...]) now adds each opening as a FreeContour feature on the plate.
- Panel.from_outlines(openings=[...]) now creates Opening panel features instead of passing raw polylines into the geometry object.
- New Opening / OpeningType classes in compas_timber.panel_features implement the full feature interface: the opening is built as a lofted Brep and subtracted from the panel geometry at feature-application time.
- New recognize_doors and horizontal_openings flags on Panel.from_outlines() enable automatic extraction of door cutouts from L-shaped wall outlines, via the new extract_door_openings() helper.
- New utility functions extend_line_segments(), get_interior_corner_indices(), and get_interior_segment_indices() support the door-detection logic.
- 
**Breaking changes:**

- Code that passes openings= to PlateGeometry(...), Plate(...), or Panel(...) directly will need to be updated to use from_outlines(openings=[...]) or add features explicitly. Code that reads plate.plate_geometry.openings will need to inspect plate._features instead.

- [x] Refactoring

### Checklist


src/compas_timber/panel_features/panel_features.py
chenkasirer: Not a change in this PR, but this should be an `StrEnum` rather than a class with class attributes

src/compas_timber/panel_features/opening.py
```diff
@@ -0,0 +1,146 @@
...
+        else:
+            for pt in outline.points:
+                pts_a.append(Point(pt[0], pt[1], 0))  # project to panel.planes[0]/panel.outline_a
+                pts_b.append(Point(pt[0], pt[1], panel.thickness))
+        pl_a = Polyline(pts_a).transformed(Transformation.from_frame(frame).inverse())
+        pl_b = Polyline(pts_b).transformed(Transformation.from_frame(frame).inverse())
+        return cls(frame, pl_a, pl_b, opening_type=opening_type, name=name)
+
+
+class OpeningType(object):
```
chenkasirer: make `StrEnum`
```
from enum import auto
from compas_timber.utils import StrEnum

class OpeningType(StrEnum):
    DOOR = auto()
    ...   
```

src/compas_timber/elements/panel.py
```diff
@@ -433,3 +414,120 @@ def from_brep(cls, brep: Brep, **kwargs):
...
+            Additional keyword arguments to be passed to the constructor.
+
+        Returns
+        -------
+        :class:`~compas_timber.elements.Panel`
+            A PlateGeometry object representing the plate geometry with the given outlines.
+        """
+
+        if openings:
+            openings = [(o, "window") for o in openings]
```
chenkasirer: why not use the `OpeningType` Enum here?

src/compas_timber/elements/panel.py
```diff
@@ -433,3 +414,120 @@ def from_brep(cls, brep: Brep, **kwargs):
...
+    """
+    combine_parallel_segments(outline_a)
+    combine_parallel_segments(outline_b)
+
+    interior_indices_a = get_interior_segment_indices(outline_a)
+    interior_indices_b = get_interior_segment_indices(outline_b)
+    interior_indices = set(interior_indices_a) | set(interior_indices_b)
+
+    openings = []
+    while True:
```
chenkasirer: I'd use an `is_door_found` flag here instead of while `True`. more explicit.

src/compas_timber/utils/__init__.py
```diff
@@ -792,6 +794,43 @@ def get_polyline_normal_vector(polyline: Polyline, normal_direction: Optional[Ve
...
+
+def get_interior_segment_indices(polyline):
+    """Get the indices of the interior segments of a polyline."""
+    interior_corner_indices = get_interior_corner_indices(polyline)
+    edge_count = len(polyline.points) - 1
+    _interior_segment_indices = []
+    for i in range(edge_count):
+        if i in interior_corner_indices and (i + 1) % edge_count in interior_corner_indices:
+            _interior_segment_indices.append(i)
+    return _interior_segment_indices
```
chenkasirer: perhaps move these to `compas_timber.geometry`? also please add typing information and a complete docstring

src/compas_timber/utils/__init__.py
```diff
@@ -792,6 +794,43 @@ def get_polyline_normal_vector(polyline: Polyline, normal_direction: Optional[Ve
...
+        ints = intersection_line_line(segments[i - 1], segments[i])
+        if not ints[0]:
+            continue
+        segments[i - 1] = Line(segments[i - 1].start, ints[0])
+        segments[i] = Line(ints[0], segments[i].end)
+
+
+def get_interior_corner_indices(polyline):
+    """Get the indices of the interior corners of a polyline."""
+    _interior_corner_indices = []
```
chenkasirer: what's with the leading underscores?

src/compas_timber/panel_features/opening.py
```diff
@@ -0,0 +1,146 @@
...
+            raise FeatureApplicationError(panel_geometry, self.shape, f"Failed to apply opening to panel geometry: {e}")
+
+    @classmethod
+    def from_outline_panel(cls, outline, panel, opening_type=None, project_horizontal=False, name=None):
+        """Creates an opening from a single outline and a panel.
+
+        The outline defined locally relative to the panel frame. The outline is projected
+        onto the outline_a side of the panel.
+
+        Parameters
```
chenkasirer, line 94: add missing parameters in docstring

src/compas_timber/elements/panel.py
```diff
@@ -433,3 +414,120 @@ def from_brep(cls, brep: Brep, **kwargs):
...
+            if door_openings:
+                if openings is None:
+                    openings = [(o, "door") for o in door_openings]
+                else:
+                    openings.extend([(o, "door") for o in door_openings])
+
+        args = PlateGeometry.get_args_from_outlines(outline_a, outline_b)
+        PlateGeometry._check_outlines(args["local_outline_a"], args["local_outline_b"])
+        kwargs.update(args)
+        panel = cls(**kwargs)
```
chenkasirer: this part seems odd: having `PlateGeometry` generate the arguments to create `Panel` and then just pass them through. if `PlateGeometry` knows how to create a `Panel` perhaps it should just create a `Panel`..

chenkasirer (changes requested):
Looks like a good direction. left a few comments. also the build fails and there are some linting issues.

src/compas_timber/utils/__init__.py
obucklin: I agree, but then I would say almost everything should go to `compas_timber.geometry`, so maybe in another PR?

src/compas_timber/elements/panel.py
obucklin: Would you do this for `Plate`, too? Layer will probably also need this. That is anyways a staticmethod, so would it make more sense to make it a function instead?

src/compas_timber/utils/__init__.py
obucklin: added issue #749

chenkasirer: yup that's fine

src/compas_timber/elements/panel.py
chenkasirer: how about we just create a `PlateGeometry` instance here and give it to `Plate`/`Panel`? they create one anyways from the data the `get_args_from_outlines` produces. and then also we can just serialize the `PlateGeometry` instance directly as part of `Plate`/`Panel` instead of implicitly serialized its data as part of these two.

obucklin: @chenkasirer I don't mind this, but then we don't have the common `frame, length, width, height` interface for `Plate.__init__`, which was a bit of an issue when it was debuted. these would of course still be properties of Plate, Panel, Layer. Not sure how to solve this in a way that doesn't break that. I could use some module level functions, eg `get_polylines_in_xy_frame` , `BoundingBox.from_points`, `get_frame_from_BBox` , `get_dims_from_BBox`... 

EDIT:
I guess we could make the frame, dimensions and `PlateGeometry` object optional arguments. If no dimensions are given then those are taken from the `PlateGeometry`. Some parsing would be required in `__init__`. If there is no `PlateGeometry` passed, then I would generate one in the `__init__` and use that for serialization and as the "truth" for properties, etc.

what do you think?

obucklin: > EDIT: I guess we could make the frame, dimensions and `PlateGeometry` object optional arguments. If no dimensions are given then those are taken from the `PlateGeometry`. Some parsing would be required in `__init__`. If there is no `PlateGeometry` passed, then I would generate one in the `__init__` and use that for serialization and as the "truth" for properties, etc.
> 

I went ahead and did it this way. let me know what you think.

src/compas_timber/panel_features/opening.py
obucklin: not sure what you mean.

obucklin:
@chenkasirer ready for another look. Yes, I will squash when I merge :-)

chenkasirer (approved):
LGTM!

PR #753, opened 2026-06-16
better tolerance isolation accross unittests

chenkasirer:
Fixed flaky tests cause by some tests that mutated the global TOL singleton (e.g. `TOL.absolute = 0.001`). the values leaked into subsequent tests, since TOL is shared global state and the per-test `reset_tolerance` fixture only reset before each test, not after.

- Added a session-scoped tolerance_session_default fixture in conftest.py to set the default tolerance before the very first test runs — module-scoped fixtures built earlier in a run weren't covered by the function-scoped reset.
- Extended reset_tolerance to also reset TOL after each test (via yield), closing the gap where a test's mutation could bleed into whatever ran next.
- Replaced direct, persistent mutation of TOL in test_slot.py and test_utils.py with TOL.temporary(...) context managers, scoping the non-default tolerance to just the code that needs it instead of relying on global resets.

### Checklist

### DIFF
--- a/tests/compas_timber/conftest.py
+++ b/tests/compas_timber/conftest.py
@@ -2,13 +2,25 @@ from compas.tolerance import TOL
 import pytest
 
 
+@pytest.fixture(autouse=True, scope="session")
+def tolerance_session_default():
+    # Session scope is broader than module scope, so this runs before any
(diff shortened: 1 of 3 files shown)


obucklin (approved):
LGTM

PR #755, opened 2026-06-18
Add orientation to plate geo

obucklin:
- Adds an orientation parameter to all alternate constructors of Panel, Plate, and PlateGeometry.from_global_outlines. When provided, the vector is projected onto the element's plane and used to override the local coordinate frame that would otherwise be derived automatically from the input outline geometry.
  - Fixes a bug in PlateGeometry.from_global_outlines where the frame-flip check (which ensures outline_b lands in the +Z half of the local frame) was evaluated after computing transform_to_world_xy. This produced a wrong transform and malformed local outlines for any outline whose natural frame normal pointed in the −Z direction, causing a ValueError: outline_a must lie on the XY-plane for a wide class of valid inputs.
  - Replaces the hard-coded outline_a[-2] third-point lookup in from_global_outlines with a backwards search for the first non-colinear point, making frame construction robust against outlines where the last segments are colinear with the first edge.

  Changed files

- `src/compas_timber/elements/plate_geometry.py`:  `orientation `parameter + robust non-colinear search + frame-flip ordering fix                       
- ` src/compas_timber/elements/panel.py `:  `orientation `threaded through `from_outlines`, `from_outline_thickness`, `from_face_thickness`, `from_brep `
-  `src/compas_timber/elements/plate.py` :         Same as `Panel`
- `tests/compas_timber/test_panel.py`:  4 new tests for `orientation ` (invariants + explicit frame assertions for flat and sloped panels)   
- `tests/compas_timber/test_plate.py`: 4 equivalent new tests for Plate                                               
- `CHANGELOG.md`:  Unreleased section updated                                                                  

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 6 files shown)


chenkasirer (approved):
LGTM

PR #761, opened 2026-06-18
add composite joint

### LINKED ISSUE #638: think about a concept to using existing joint implementations in other ones
context is: https://github.com/gramaziokohler/compas_timber/pull/619#discussion_r2598013506

### LINKED ISSUE #687: Add support for K/Y joints with topology rules in Grasshopper

### LINKED ISSUE #746: composable joints
for PanelJoints with multiple layers and for Beam joints with 3+ elements, we should consider a mechanism to compose their joints out of multiple pairwise joints.

obucklin:
This introduces the `CompositeJoint`, which is composed of a list of pairwise joints. It is intended to serve as a simpler interface for generating 3+ element joints without implementing specific joints for each case. This is developed in conjunction with a `CompositeJointRule` in timber_design, which allows different `JointRule` sets to be implemented on `Clusters` than for "clean" pairwise joints. 

[CompositeJointRule](https://github.com/gramaziokohler/timber_design/pull/13/)

Essentially, individual elements are still acted upon by pairwise joints, but this `CompositeJoint` is what gets registered on the TimberModel, allowing us to distinguish between purely pairwise joints and those based on higher-order logic. 

The concrete reason for this change is in the PanelPopulator. Specifically with the edge beams, I want to define one joining behavior when they meet "cleanly", however, when a stud beam hits the intersection between two edge beams, this will often result in janky feature application. With this I can apply different joining parameters in those cases. 

I'd love to hear your feedback and see what you all think. 

### Checklist


obucklin:
I added a couple of examples here, both as python and GH script: `timber_design\examples\composite_joint_rule\` on the `timber_design\add_composite_rule` branch.
@papachap @chenkasirer would you guys please take a look at this?

obucklin:
@papachap I implemented the changes we discussed. please take a look

src/compas_timber/connections/composite_joint.py
```diff
@@ -0,0 +1,128 @@
...
+        The approximate location of the joint, taken from the `cluster.location`.
+    topology : :class:`~compas_timber.connections.JointTopology`
+        The topology of the joint, taken from the `cluster.topology`.
+    """
+
+    SUPPORTED_TOPOLOGY = JointTopology.TOPO_UNKNOWN
+    MIN_ELEMENT_COUNT = 3
+    MAX_ELEMENT_COUNT = None
+
+    def __init__(self, joints, name=None, cluster=None, **kwargs):
```
papachap: I'd avoid passing the cluster here as it becomes live joint state, but `__data__` never serializes it. Even though location/topology do round-trip without it, it feels a bit implicit.

papachap: Better to make topology/location explicit params on `__init__` (defaulting to values computed from joints if not passed), snapshot them once like `promote_joint_candidate` does with the candidate, and not hold onto cluster at all.

src/compas_timber/connections/cluster_joint.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @property
+    def location(self):
+        """The approximate location of the joint, taken from the first sub-joint."""
+        return self.cluster.location
+
+    @property
+    def topology(self):
+        """Returns the topology of the joint if there is only one joint, otherwise TOPO_UNKNOWN."""
+        return self.cluster.topology
```
papachap: Here I'd go with something like this, even though it does bring a bit of code duplication. 
Building and discarding a `Cluster` instance just to read .location/.topology off it feels a bit like a smell.

```suggestion
    @property
    def location(self) -> Point:
        return self.location_of(self.joints)

    @staticmethod
    def location_of(joints) -> Point:
        return sum([j.location for j in joints[1:]], joints[0].location) / len(joints)

    @property
    def topology(self):
        return self.topology_of(self.joints)

    @staticmethod
    def topology_of(joints):
        if len(joints) == 0:
            return JointTopology.TOPO_UNKNOWN
        if len(joints) == 1:
            return joints[0].topology
        if any(j.topology not in [JointTopology.TOPO_L, JointTopology.TOPO_I, JointTopology.TOPO_T, JointTopology.TOPO_X] for j in joints):
            return JointTopology.TOPO_UNKNOWN
        if any(j.topology in (JointTopology.TOPO_T, JointTopology.TOPO_X) for j in joints):
            return JointTopology.TOPO_K
        return JointTopology.TOPO_Y
```

src/compas_timber/connections/cluster_joint.py
```diff
@@ -0,0 +1,128 @@
...
+            The pairwise sub-joints.
+
+        Returns
+        -------
+        :class:`~compas_timber.connections.CompositeJoint`
+        """
+        joint = cls(joints=joints, **kwargs)
+        model.add_joint(joint)
+        return joint
+
```
papachap: `CompositeJoint` can't actually go through `promote_cluster` mechanism as is. The base `Joint.promote_cluster` calls `cls.create(model, *elements, **kwargs)`, but `CompositeJoint.create` expects other stuff. 

I think this needs a `promote_cluster` override that would pull topology/location off the cluster the same way as `promote_joint_candidate`. This links back to my comment of passing topology/location params to `__init__`.

papachap (changes requested):
Hey @obucklin, this looks good overall.. left some comments, thoughts and suggestions.

src/compas_timber/connections/composite_joint.py
obucklin: Hey I very much disagree. I would even propose we change this to `ClusterJoint`. A cluster is just a collection of pairwise joints, which is exactly what `CompositeJoint`/`ClusterJoint` takes as main argument. I think we should just pass a cluster so we can avoid possible conflicts between `joints` and `cluster.joints`

src/compas_timber/connections/cluster_joint.py
obucklin: See my above comment

src/compas_timber/connections/cluster_joint.py
obucklin: `promote_cluster` is not called anywhere in CT or timber_design. Are you actually using it. I will override this and include some similar checks to `Joint.promote_cluster`

obucklin: @papachap check out #802

obucklin:
@chenkasirer hey can you please take a look at this and #802? I've been waiting for over a month.

chenkasirer (approved):
LGTM

src/compas_timber/connections/cluster_joint.py
obucklin: @papachap can you take a look again please?

chenkasirer:
@obucklin @papachap 

Just my two cents on the discussion. I think 4 classes got a bit entangled in one another in this PR (or maybe a bit earlier), and these are `Joint`, `JointCandidate`, `Cluster` and now `ClusterJoint`.

I think `Cluster` is the odd-one-out here. It was initially a transient container for `JointCandidates` that formed a group based on its geometry, a helper for creating proper concrete joinery downstream. is was created by `get_clusters_from_joint_candidates()` and consumed by `Joint.promote_cluster()`, that's it.

Now it assumes a second role which is has a different lifecycle: it composes REAL joints and saved in the model and in the graph (as part of `ClusterJoint`).

This makes `Cluster` do two jobs instead of one and makes `ClusterJoint` seem like a thin wrapper around it.

So I'd propose the following:
1. keep `Cluster` transient container of `JointCandidate` - AKA no changes in this PR.
2. have `ClusterJoint` take list list sub-joints. `ClusterJoint.promote_cluster()` could do something much closer to what the vanilla one does: promote the candidates in a `Cluster` to real joints and then add THOSE to the `ClusterJoint` along with the topology and location and anything else it might need - given this role, the previous name `CompositeJoint` arguably better describes what this class does.

Finally, I think we should consider forming a more explicit separation between `JointCandidate` and `Joint`. the former being the latter causes a lot of confusion and I think we'd be better off treating these two as the two different concepts they are.

obucklin:
@chenkasirer hey thanks for the reply. I agree with most of your comments. I have no problem reverting to `CompositeJoint` and leaving `Cluster` untouched. 

My only point of contention is with the `promote_cluster` method. The point of this joint is to keep the contained pairwise joints together in a single joint instance in the `TimberModel`, letting the user/model know that "these joints were created together and depend on each other". The logic normally implemented in `Joint.promote_cluster` - the parsing of joint_candidates and relationships between elements - is inherently external to this joint, and is handled in the `ClusterRule`/`CompositeRule`/`JointRuleSolver`(JRS). If we want to implement `promote_cluster` here, then all of that parsing logic must also be implemented/passed in. If the `Cluster` is _only_ a container for `JointCandidate`s, then I don't see any sense in implementing `promote_cluster` here. 

Another option would be to compose this joint with a colleciton of `JointRule`s. 
**Pros:**
- We could internalize the `JointCandidate`/`Element` parsing and implement `promote_cluster` in a meaningful way. 
- We could implement the `create` method where we pass the joint a list of elements.
 
**Cons:**
- We would do the same parsing in two places, one in the JRS where we determine whether/which joint type matches the cluster, and the other internally in `promote_cluster` to determine how to apply the pairwise joints (which was presumably already done in JRS). Although we could have an alternate constructor that allows us to pass in the pre-assigned joints from JRS.
- We would want to bring `JointRule`s nad the JRS back to compas_timber to avoid CT depending on timber_design.

Let me know what you guys think!
@papachap you too ;-)

obucklin:
Question, does it even make sense to have a `CompositeJoint.create` method if the interface is completely different?

chenkasirer:
> @chenkasirer hey thanks for the reply. I agree with most of your comments. I have no problem reverting to `CompositeJoint` and leaving `Cluster` untouched.
>
Agreed. `CompositeJoint`, and leave `Cluster` as the transient candidate container.
 
> My only point of contention is with the `promote_cluster` method. 
... If we want to implement `promote_cluster` here, then all of that parsing logic must also be implemented/passed in. If the `Cluster` is _only_ a container for `JointCandidate`s, then I don't see any sense in implementing `promote_cluster` here.

Also agree. A shared `promote_cluster` can't really work: for a multi-beam joint it's "cluster -> one joint over all elements," while for a composite it's "cluster -> N pairwise sub-joints, type chosen per candidate externally" That per-candidate decision is the rule logic we agreed shouldn't live on the joint.

In-fact, `promote_joint_candidate` also doesn't do much; since joint candidates stay in-place, it's simply adding a Joint to the model + passthrough or location and topology which are anyways calculated in Joint.

I'd get rid of both `promote_*` methods and leave it to external design logic to use the existing mechanisms i.e. `Joint.create(model, elements, ...)`/`CompositeJoint.from_joints(model, subjoints, ...)` (overriding `create` with a different signature is also a bad idea)

I think that a `CompositeJoint` that accepts a list of concrete joints and takes care of the representation of composite joints in the model as well as all the required mechanics is plenty powerful for CT. 

The JointRules system is an opinionated design interface that lives in timber design and shouldn't leak into CT which provides the building blocks. The product of the CompositeRule should simply be a CompositeJoint.

obucklin:
@papachap Hey @chenkasirer and I are in agreement about this, and I implemented all the changes. I deviated in 2 minor details:
- I extracted the `Cluster.topology` solving to a module-level function `cluster.get_topology_from_joints(joints)`, which I call in both `Cluster.topology` and `CompositeJoint.topology`
- I didn't implement the implied `CompositeJoint.create()` replacement, `CompoisteJoint.from_joints()` as that is the default constructor anyways.

Please approve if you don't have anything else to add. Thanks!

tests/compas_timber/test_composite_joint.py
```diff
@@ -0,0 +1,370 @@
...
+
+
+# ---------------------------------------------------------------------------
+# 5. Location
+# ---------------------------------------------------------------------------
+
+
+def test_location_delegated_to_first_sub_joint(composite, sub_joints):
+    """CompositeJoint.location falls back to first sub-joint's location when not set."""
+    assert composite.location is not None
```
papachap, line 246: just a leftover from earlier
```suggestion
def test_location_is_average_of_sub_joint_locations(composite, sub_joints):
    """CompositeJoint.location is the average of its sub-joints' locations, not just the first."""
    expected = sum((j.location for j in sub_joints[1:]), sub_joints[0].location) / len(sub_joints)
    assert composite.location == expected
```

papachap (approved):
LGTM! Thanks @obucklin for the adjustments.

Just one thing i would propose is to add a `promote_cluster` override in `CompositeJoint` that raises `NotImplementedError` (pointing to `CompositeJoint.create()` instead), until the `promote_*()` methods get removed (maybe worth opening a ticket for it).

PR #754, opened 2026-06-16
fixed t butt joint cutting cross beam

chenkasirer:
* Fixed `TButtJoint` errorneously cutting cross beam even though `modify_cross` is set to `False`.
* Fix TButtJoint cutting/extending the cross beam under certain conditions, by moving modify_cross/back_plane out of the shared ButtJoint base into LButtJoint, where they're actually used.
- TButtJoint no longer has any trace of modify_cross (no attribute, constructor param, or serialized field) — it's structurally impossible for a T-Butt to modify the cross beam now, rather than relying on a hardcoded override.
 
### What type of change is this?

- [x] Bug fix in a **backwards-compatible** manner.
- [ ] New feature in a **backwards-compatible** manner.
- [ ] Breaking change: bug fix or new feature that involve incompatible API changes.
- [ ] Other (e.g. doc update, configuration, etc)

### Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [ ] I ran all tests on my computer and it's all green (i.e. `invoke test`).
- [ ] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [ ] I added new functions/classes and made them available on a second-level import, e.g. `compas_timber.datastructures.Beam`.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added necessary documentation, including updating `class_diagrams.rst` (if appropriate).

PR #762, opened 2026-06-18
Suggestions

chenkasirer:
Suggestions for https://github.com/gramaziokohler/compas_timber/pull/737

* reapplied a fix imlemented in https://github.com/gramaziokohler/compas_timber/pull/754 and removed in this PR
* consollidated the plane spec scalars on butt and miter joints  into two new classes: `CutPlaneSpec` and `MiterPlaneSpec`.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Added `CutPlaneSpec` to `compas_timber.connections` — a serializable, beam-relative description of a cutting plane stored as `(ref_side_index, angle, offset)`. Build one from a world-coordinate `Plane` with `CutPlaneSpec.from_butt_plane()` or `CutPlaneSpec.from_back_plane()`; resolve it back to a `Plane` with `.to_plane(beam)`.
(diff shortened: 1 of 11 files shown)


obucklin (changes requested):
@chenkasirer hey I appreciate the suggestions. I think the `modify_cross` / apply `L*Joint` to a TOPO_T / apply `T*Joint` to a TOPO_L, needs some consideration, but I'm fine to leave as you change it here. 

I like how clean the `*PlaneSpec` is, but this will not work with the GH workflow, as the GH component params are taken from the joint's `__init__`. We can certainly discuss if we want to change that workflow or add `*PlaneSpec` GH components. At the least I would suggest changing the arg name to `miter_plane_spec` instead of `miter_plane` to make it clear that it doesn't take a `Plane` object. I also think if you are making a new class for this, maybe we should have a single class instead of separate `MiterPlaneSpec` and `JointCutPlane` classes.

chenkasirer:
@obucklin I strongly feel that under no circumstance should a TButt behave like an LButt, that's what the LButt is for. forcing TButt on beams meeting as an L or vice-versa can be done by doing just that.

I agree with the name change, will do.

regarding the compatibility with the GH component, I see the issue, I think adding a plane spec component it legit, or building some custom logic for these joints at the GH end. either way, i think the GH workflow should bend to the library and not the other way around.

regarding making this a single class, I've considered and actually preferred this initially, however the two aren't quite compatible with each other and it would be imho forced to use a single class with potentially incompatible argument values. I'd rather have an extra class than have to enforce the right combination of argument values in various places.

obucklin:
Hey @chenkasirer thanks for the chat today. I'll go ahead and merge

PR #776, opened 2026-07-07
FrenchRidgeLap: fix wrong RefPosition for 90° L-joints caused by floating-point comparison

### LINKED ISSUE #774: LFrenchRidgeLap Bug
**Describe the bug**

the generated rhino/gh geometry and the btlx file of the joint is wrong. When joining at beam_a start point and beam_b end point OR beam_a end point and beam_b start point. one of the angle of one of the beam is not correct.

**To Reproduce**

1. Context

Image

Rhino: 2 lines: cmain: x positive, cother: y positive
GH: python component. inputs: cmain (line, item access), cother (line, item access)
2. Sample script
gh component script attached in the FRL_bug.py file

3. Sample data

**Expected behavior**

one of the angle of one of the beam if off.

**Screenshots**

Image

**Desktop (please complete the following information):**

- OS: Windows 11
- Python version 3.13.14
- Python package manager pip and conda_forge
- compas_timber version 2.1.2-71e9aac7

**Additional context**

paulocinco:
Problem
For certain [LFrenchRidgeLapJoint] configurations at exactly 90°, one of the two beams received the wrong RefPosition in its [FrenchRidgeLap] feature (refedge instead of oppedge or vice versa). This corrupted both the exported BTLx file and the 3D geometry rendered in Rhino/Grasshopper, since the ridge pocket was machined on the wrong edge of the beam.

The bug was triggered only when a beam's centerline ran in a direction that introduced floating-point drift in the angle calculation (e.g. a beam defined from (2000, 0, 0) to (0, 0, 0) instead of (-2000, 0, 0) to (0, 0, 0)). In that case [_calculate_angle] returned 90.000000000000014° rather than exactly 90.0°.

Root Cause
[_calculate_ref_position] in [french_ridge_lap.py] had a special branch for orthogonal connections gated on if angle == 90.0:. The exact float equality test failed when the angle differed from 90.0 by ~1.4 × 10⁻¹⁴ degrees. This caused the general [angle > 90.0] branch to be used instead, which applied the wrong [is_ref_edge] condition and returned the wrong RefPosition.

There was also a stray print(90) debug statement inside that branch that has been removed.

Fix
Replace the exact equality [angle == 90.0] with [TOL.is_close(angle, 90.0)] (consistent with how tolerance checks are done elsewhere in the codebase, e.g. [ISimpleScarf]). The same guard is applied to the [is_ref_edge] condition below.

Testing
All 30 existing tests pass. The fix was validated by confirming that two physically identical 90° L-joints — differing only in whether the beam centerline runs in the +X or −X direction — now produce identical RefPosition values.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
-
+* Fixed wrong `RefPosition` assigned to one beam in `LFrenchRidgeLapJoint` for 90° configurations where floating-point drift caused `_calculate_ref_position` to miss the orthogonal-connection branch (`angle == 90.0` replaced with `TOL.is_close(angle, 90.0)`). Also removed a stray `print(90)` debug statement.
 ### Removed
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,8 @@ authors = [
     { name = "Beverly Lytle", email = "lytle@arch.ethz.ch" },
     { name = "Eric Gozzi", email = "eric.gozzi@arch.ethz.ch" },
     { name = "Rodrigo Arca Zimmermann", email = "rodrigo.arca@gmail.com" },
-    { name = "Nicolas Benjamin Boscoboinik", email = "boscoboinik@arch.ethz.ch" }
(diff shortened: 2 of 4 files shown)


obucklin (approved):
Looks LGTM.

PR #780, opened 2026-07-09
Fix dev-setup instructions: requirements-dev.txt no longer exists

daniel-locatelli:
## What

`CONTRIBUTING.md` still tells contributors to install dev dependencies with `pip install -r requirements-dev.txt`, but the requirements files were removed when dependency declarations moved into `pyproject.toml` (+ `uv.lock`) — so the documented setup fails on a fresh clone.

- **CONTRIBUTING.md**: point to `uv sync --extra dev` (uses the committed lockfile) as the primary path, with `pip install -e ".[dev]"` as the tool-agnostic alternative.
- **pyproject.toml**: remove the `requirements.txt` / `requirements-dev.txt` references from `[tool.setuptools.dynamic]` — both files are gone, and the entries are ignored anyway since `dependencies` is declared statically in `[project]`.
- **CHANGELOG.md**: entries under Unreleased.

## How verified

Fresh clone on a clean Windows 11 machine: `uv sync --extra dev` → editable install resolves in ~20 s; after the `pyproject.toml` change, forced rebuild of the editable install + `uv run pytest` → **822 passed**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `TimberModel.remove_joint()` now calls `Joint.reset_location()`.
 * Fixed a bug in `PlateGeometry.from_global_outlines` where the frame-flip was applied incorrectly when the initial local frame's normal pointed in the −Z direction.
 * Bumped minimum required `compas_brep` due to bugfix in Grasshopper Brep scene object.
+* Fixed development setup instructions in `CONTRIBUTING.md`: `requirements-dev.txt` no longer exists — use `uv sync --extra dev` or `pip install -e ".[dev]"`.
+* Removed stale `requirements.txt` / `requirements-dev.txt` references from `[tool.setuptools.dynamic]` in `pyproject.toml` (both files are gone; dependencies are declared statically in `[project]`).
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `TimberModel.remove_joint()` now calls `Joint.reset_location()`.
 * Fixed a bug in `PlateGeometry.from_global_outlines` where the frame-flip was applied incorrectly when the initial local frame's normal pointed in the −Z direction.
 * Bumped minimum required `compas_brep` due to bugfix in Grasshopper Brep scene object.
+* Fixed development setup instructions in `CONTRIBUTING.md`: `requirements-dev.txt` no longer exists — use `uv sync --extra dev` or `pip install -e ".[dev]"`.
+* Removed stale `requirements.txt` / `requirements-dev.txt` references from `[tool.setuptools.dynamic]` in `pyproject.toml` (both files are gone; dependencies are declared statically in `[project]`).
```
chenkasirer: changes to project files and documentation usually don't require a changelog entry. the changelog more for developers using the library and used to announce e.g. API changes etc.

chenkasirer (approved):
LGTM! just a small comment

PR #782, opened 2026-07-09
Add Python 3.12-3.14 trove classifiers

daniel-locatelli:
## What

Adds `Programming Language :: Python :: 3.12 / 3.13 / 3.14` trove classifiers to `pyproject.toml` (plus a changelog entry).

## Why

The CI matrix in `.github/workflows/build.yml` tests against Python 3.9, 3.12 and 3.14 on every push, but the classifiers published to PyPI still advertise only 3.9–3.11. Anyone checking supported versions on the [PyPI page](https://pypi.org/project/compas-timber/) (or filtering by classifier) gets stale information.

`requires-python = ">=3.9"` is untouched — this only updates the advertised upper range to match what CI actually verifies.


chenkasirer (approved):
Nice, thanks! 
Just one comment: changes like these (project files, documentation, etc) don't need a changelog entry. 
you can assign the PR with the label `no changelog` so that the PR checker doesn't complain.

daniel-locatelli:
Good to know, thanks!

PR #795, opened 2026-07-14
Centralize joint feature/extension lifecycle in TimberModel

### LINKED ISSUE #783: When a joint is added to TimberModel over an existing joint, the original joint's features are not removed from the elements.
Possible solutions:
1. `TimberModel.add_joint()` checks for Joint already on that edge and tells it to remove its features before new joint is added to model. 
2. Joint creates features and stores them on graph edges directly. This would use the pair of directional edges to indicate which features apply to which elements, similar to `compas_model.modifiers`. Joint.features would reference the features on the edge it occupies. `Joint.add_features()` already removes its own features before adding new features. we could use this to get rid of the `is_joinery` flag on BTLxProcessing.

### LINKED ISSUE #792: Add guard against multiple calls to process_joinery()
process_joinery() has no check to prevent it being called more than once on the same model, which can cause issues

obucklin:
What type of change is this?

- [x] Bug fix in a backwards-compatible manner.
- [x] New feature in a backwards-compatible manner.

Summary

Makes Joint.features the single source of truth for what a joint has applied to its elements, and moves responsibility for clearing that state from each joint (inconsistently, it turned out) to TimberModel:

- Joint.__init__ now initializes self.features = []; Joint.clear_features()/Joint.clear_extensions() are new base methods that undo everything a joint has applied.
- TimberModel.remove_joint() now calls both before tearing a joint down, instead of leaving features/extensions stuck on the remaining elements.
- TimberModel.add_joint() now removes any existing joint between the same elements first (using TimberModel.get_joint(), new), preventing a stale joint from becoming unreachable while still holding features on the beams.
- TimberModel.remove_element() now cascades into remove_joint() for every joint touching the removed element (previously it relied on `compas_model.Model.remove_element`, which removed edges/interactions connected to the element, but not the `compas_timber.Joint` from the `TimberModel._joints` dict.
- TimberModel.process_joinery() gained an optional joints_to_process parameter and now clears every processed joint's features up front, so individual joints no longer need self.clear_features()/a private _clear_features() call at the top of their own add_features().
- Every joint's add_features() was audited and updated to actually put everything it applies into self.features (previously some subclasses silently dropped part of it).

Why this approach

Per-joint self-clearing was the status quo before this branch, and it was quietly wrong in three places: BallNodeJoint never recorded any feature in self.features at all; YButtJoint overwrote self.features at the end of add_features() instead of extending it, dropping its main cut and mill-depth pocket features from tracking; TOliGinaJoint applied three text-engraving features that were never added to self.features. In all three cases, clearing the joint would silently leave those features stuck on the beam forever. Centralizing the clear step in TimberModel and requiring every joint to fully report what it applied removes this whole class of bug, at the cost of touching many joint files mechanically (removing now-redundant self.features = [] inits and inline self-clear blocks).

Notable fixes bundled in

- `Beam.remove_blank_extension()` used `del self._blank_extensions[joint_guid]`, which raised KeyError for any joint that didn't extend every one of its elements (e.g. ButtJoint only extends main_beam) — this broke remove_joint() for the majority of joint types. Now uses `.pop(joint_guid, None)`, matching remove_features()'s existing tolerant pattern. (Reproduced directly: 29 tests failed with this exact KeyError before the fix.)
- PlateJoint.clear_extensions() called remove_blank_extension(None) on a plate's cross element whenever the joint's b_segment_index was None (e.g. PlateTButtJoint, which never sets an extension on its cross plate) — remove_blank_extension(None) resets every edge extension on that plate, silently wiping out unrelated joints' extensions on the same element. Now guards both sides on is not None.
- Renamed Beam.add_blank_extension/remove_blank_extension's `joint_key` parameter to `joint_guid` (it's always a GUID, never a key).
- `TimberModel.process_joinery` is now _**idempotent.**_

### Checklist


src/compas_timber/model.py
```diff
@@ -343,6 +366,15 @@ def add_joint(self, joint):
         joint : :class:`~compas_timber.connections.joint`
             An instance of a Joint class.
         """
+        # explicitly remove old joint instances between the same elements.
```
chenkasirer, line 388: worth mentioning this in the docstring so that a user understands it does that

src/compas_timber/elements/beam.py
```diff
@@ -324,41 +324,41 @@ def from_box(cls, box, **kwargs):
 
     @reset_computed
     @reset_timber_attrs
-    def add_blank_extension(self, start, end, joint_key=None):
+    def add_blank_extension(self, start, end, joint_guid=None):
```
chenkasirer: this is a breaking API change, the PR is marked as backwards compatible. please consider if this is really necessary and if so update the PR checklist

chenkasirer (approved):
there's a couple breaking API changes in this PR (argument renaming) so for transparency's sake I'd suggest flagging it as such. otherwise LGTM

src/compas_timber/elements/beam.py
```diff
@@ -324,41 +324,41 @@ def from_box(cls, box, **kwargs):
...
         # type: (float, float, None | int) -> None
         """Adds a blank extension to the beam.
 
         start : float
             The amount by which the start of the beam should be extended.
         end : float
             The amount by which the end of the beam should be extended.
-        joint_key : int
+        joint_guid : int
```
papachap:
```suggestion
        joint_guid : str
```

src/compas_timber/elements/beam.py
```diff
@@ -324,41 +324,41 @@ def from_box(cls, box, **kwargs):
...
     @reset_timber_attrs
-    def remove_blank_extension(self, joint_key=None):
+    def remove_blank_extension(self, joint_guid=None):
         # type: (None | int) -> None
         """Removes a blank extension from the beam.
 
         Parameters
         ----------
-        joint_key : int
+        joint_guid : int
```
papachap:
```suggestion
        joint_guid : str
```

papachap (approved):
LGTM too!

src/compas_timber/elements/beam.py
obucklin: reverted to joint_key

src/compas_timber/elements/beam.py
obucklin: as per @chenkasirer s comments I reverted to `joint_key`

src/compas_timber/elements/beam.py
obucklin: see previous response

PR #796, opened 2026-07-14
Normalize FeatureApplicationError/BeamJoiningError geometry to model space

### LINKED ISSUE #779: Inconsistent application of local and global geometry in Joining and FeatureApplicationErrors
The geometry one gets in the BeamJoiningErrors and FeatureApplicationErrors is sometimes in local beam space and sometimes in global space. This makes it very difficult to identify what and where the issue is. 

In my opinion this should always be in global space to help identify the problematic elements and joints in situ.

obucklin:
FeatureApplicationError and BeamJoiningError carry geometry meant for downstream error visualization (the GH DebugInfomation component draws it directly in the Rhino document — i.e. world/model space). The geometry space was inconsistent depending on where the error was raised:

- FeatureApplicationError: errors raised inside `BTLxProcessing.apply()` (~55 sites across 16 fabrication/*.py files) stored local/element-space geometry, since that's the space apply() itself operates in. Errors raised outside apply() (`Drilling.from_line_and_element`, `BTLxFromGeometryDefinition.feature_from_element`) already used global/model-space geometry. Same exception type, two conventions depending on call path.
- BeamJoiningError: mostly already consistent, but 3 call sites misused the constructor's (beams, joint, debug_info=None, debug_geometries=None) signature — one (plate_joint.py) would actually crash with TypeError rather than raise BeamJoiningError, since it omitted the required joint argument.

This PR:
1. Makes every FeatureApplicationError raised inside apply() transform its geometry to model space inline before construction (`geometry.transformed(beam.modeltransformation)`), matching the already-global sites.
2. Fixes the 3 BeamJoiningError constructor bugs found while auditing every call site for the same consistency issue.
3. Documents the "always model space" invariant on both exception classes' docstrings, and makes `BeamJoiningError.debug_geometries` tolerant of a single geometry being passed instead of a list.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `brep_from_outlines` to `compas_timber.geometry`.
 
 ### Changed
+* `FeatureApplicationError` raised from `BTLxProcessing.apply()` now carries geometry in the model's global coordinate system (previously local/element space), matching errors raised elsewhere. 
+* Fixed a live crash (`TypeError`) and two other constructor-argument bugs on `BeamJoiningError` call sites.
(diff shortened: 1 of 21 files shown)


chenkasirer (approved):
LGTM

PR #785, opened 2026-07-10
Pr1 compute topologies

obucklin:
Refactors three near-duplicate `connect_adjacent_*` implementations with `TimberModel.compute_topologies(elements=None, max_distance=None)`, dispatching by (type_a, type_b) via `_CONNECTION_HANDLERS` — defined in `connections/solver.py` alongside the solvers it calls. Clears all joint candidates unconditionally on every call and no longer removes existing concrete joints. 

keeps all the `connect_adjacent_*` methods for backward compatibility. These now use the new `compute_topologies()` method internally.

### What type of change is this?

- [ ] Bug fix in a **backwards-compatible** manner.
- [ ] New feature in a **backwards-compatible** manner.
- [ ] Breaking change: bug fix or new feature that involve incompatible API changes.
- [x] Refactoring

### Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [x] I ran all tests on my computer and it's all green (i.e. `invoke test`).
- [x] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [ ] I added new functions/classes and made them available on a second-level import, e.g. `compas_timber.datastructures.Beam`.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added necessary documentation, including updating `class_diagrams.rst` (if appropriate).

Issue #639, opened 2025-12-10
Evaluate the coupling/dependency between: `Analyzer`/`Cluster`/`Joint`/`JointCandidate`/`JointRules`


chenkasirer:

obucklin:
I've been looking at this a bit this morning, and I have a few thoughts. 

- The `JointRuleSolver` doesn't really do any analysis of the connections. The very high level method `JointRuleSolver.apply_rules_to_model()` calls the `get_clusters_from_model()` function where the `JointCandidate`s and `Cluster`s are generated. All of the `JointRule` functionality is essentially parsing data held by `JointCandidate`s and `Cluster`s.
- On the back end, `JointCandidate` and `Cluster` do different things, namely, `JointCandidate` acts on the graph and `Cluster` proved groupings of `JointCandidate`s. On the user side, they do rather similar things, for example each has a `topology` property and would potentially be used to generate a `Joint` instance.  
- `JointCandidate` orders it's elements based on the results of `find_topology()`, which is what allows us to promote it to a `Joint` directly. `Cluster` does not have this feature, meaning we cannot do this for 3+ element clusters. 
- As an example of a workflow that uses `JointRule`s without clusters or joint candidates, look at the [this method in the slab-populator workflow](https://github.com/gramaziokohler/timber_design/blob/e2b35f5/src/timber_design/element_generators/element_generator_parameters.py#L101-120). We could definitely upstream this and create a similar function for `TopologyRule`, but this may be complicated to parse different rule types together and to handle 3+ element joints.
- One idea would be to have clusters be the main object for passing to `JointRuleSolver` and would contain all the data like elements organized by role, topology, distance, location, etc. Cluster could then have a `from_joint_candidates()` constructor, but could also take lists of elements or the full set of data.
- **summary:** The main GH/fully automated `JointRule` process requires a cluster-like object that has the required data like element count, categories, and topology. We can discuss where these are calculated and how they are passed to the `JointRuleSolver`.

Issue #511, opened 2025-08-21
Consider making JointCandidate its own thing


chenkasirer:
`JointCandidate` is a strange bird, it's technically a joint but when added to the model like a joint (via `JointCandidate.create()) it doesn't do it's job.. has to be added by using `model.add_joint_candidate` so that it ends up on the right edge attribute.

Easy fix: override `JointCandidate.create` and raise Error.

Deep fix: figure out what `JointCandidate` is, and potentially detach it from joint.

Additionally: try to consolidate the different candidate classes (`PlateJointCandidate`)

PR #801, opened 2026-07-17
Fix `solver.py`'s local-import workaround by relocating candidate dispatch

papachap:
#785 added a local import inside `solver.py`. That's new for this package: every other file in connections/ imports downward, concrete joints depend on the base files, never the other way around, and `solver.py` had no dependencies on the rest of the package until now.

The dispatch table this PR added needed that local import because it depends on `JointCandidate`/`PlateJointCandidate`, which depend back on `solver.py` through `joint.py` and `plate_joint.py`.

I think this needed adjusting, not because local imports are inherently wrong, but because of what they revealed here: the dispatch logic just doesn't belong in `solver.py`: it composes solvers, candidate classes, and element types, which makes it a layer above `solver.py`, not part of it..

The table is also meant to grow, more type combinations are already flagged as next. Each new one would need the same workaround, another local import in `solver.py`.

This PR: 
- moves that dispatch logic into its own module, `candidate_dispatch.py`, which sits above both `solver.py` and the candidate classes since it genuinely needs both, and pulls in the element types too. All of that is now a normal top-level import.

- I also cleaned up how it's consumed. The model used to call one function to find the right handler and then call that handler itself; that's now a single `get_connection_candidate(element_a, element_b, max_distance)`, so the model doesn't need to know handlers exist at all. 

- And since handlers are meant to keep being added, each one now declares the element-type pair it supports with a decorator right above its own definition, instead of a separate mapping further down the file that you'd have to cross-reference by name. The old handler-lookup function is still there since it's useful on its own for tests, but it's no longer exposed outside this module, nothing outside it calls it directly anymore now that `get_connection_candidate` is the real entry point.

### Checklist


src/compas_timber/connections/candidate_dispatch.py
```diff
@@ -0,0 +1,74 @@
...
+    result = PlateConnectionSolver().find_topology(element_a, element_b, tol=TOL.relative, max_distance=max_distance)
+    if result.topology is JointTopology.TOPO_UNKNOWN:
+        return None
+    kwargs = {"topology": result.topology, "a_segment_index": result.a_segment_index, "distance": result.distance, "location": result.location}
+    if result.topology == JointTopology.TOPO_EDGE_EDGE:
+        kwargs["b_segment_index"] = result.b_segment_index
+    return PlateJointCandidate(result.plate_a, result.plate_b, **kwargs)
+
+
+def find_connection_handler(element_a, element_b):
```
obucklin, line 63: This was a bit confusing to me at first. I think that `find_connection_handler` should just return the correct solver type so that we can call different `*ConnectionSolver` methods as necessary. Maybe put the JointCandidate creation into new `*ConnectionSolver` methods.

obucklin (changes requested):
Hey @papachap this is a good refactoring. Just one comment below. 

I am currently working on a `BeamPlateConnectionSolver` and the topologies are getting a bit complex. I am considering things like a new data class that holds topo information. Basically a refactoring/unified interface for the `*ConnectionSolver` and `*SolverResults` classes. Definitely interested to hear your thoughts.

src/compas_timber/connections/candidate_dispatch.py
```diff
@@ -0,0 +1,74 @@
...
+# Handlers register the type pair(s) they support via `@_register`, order-independent
+# (both (a, b) and (b, a) pairs match). To support a new type combination (e.g. beam-to-plate),
+# add a handler decorated with `@_register(TypeA, TypeB)` once the corresponding
+# topology-detection geometry exists.
+# ------------------------------------------------------------------
+
+_CONNECTION_HANDLERS = {}
+
+
+def _register(type_a, type_b):
```
chenkasirer, line 30: is it envisioned that new connection handlers can be registered by non-library contributors, e.g. other packages that are dependent on compas timber? if so, maybe no `_`.

chenkasirer (approved):
LGTM

src/compas_timber/connections/candidate_dispatch.py
obucklin: I have a `2dConnectionSolver` in timber_design. More thorough outline-based analysis rather than centerline. It may be something to upstream or possibly register here.

papachap: Don't have a certain opinion about that.. What do you say @obucklin?

src/compas_timber/connections/candidate_dispatch.py
papachap: I am not sure I have the full picture of what the actual pain point is.. Would the entry point of this in the model change then?

Regardless of that, this might be a bit outside of the scope of this PR, which was meant to remove the circular/lazy import issue without any rework of underlying concepts.

Happy to pick this up in a separate PR once the goal's clearer, but you're probably better placed for it given you've been hands-on with the new solver types and where the complexity actually is.

papachap:
@obucklin I replied to your comment.. I am happy to discuss your ideas! 
My opinion is that this rework might need to be done in a separate PR unless something is contradicting. LMK

src/compas_timber/connections/candidate_dispatch.py
obucklin: lets leave the connectionSolver2d out of consideration for now. I would say leave it private

obucklin (approved):
Hey regarding my previous comments about having the solvers generate the candidates themselves, I still think this is a good idea. I tried to quickly implement it, but i got all the import errors because of the JointCandidate inheritance of Joint. There are two open issues on that #511 and #639. I think that should be addressed first and then we can make those changes.

obucklin:
> @obucklin I replied to your comment.. I am happy to discuss your ideas! My opinion is that this rework might need to be done in a separate PR unless something is contradicting. LMK

@papachap Sorry maybe my last comment was unclear. I was looking at implementing my suggested changes in a fork of this branch, but there were major circular import problems due to JointCandidate inheriting Joint. I address that in another PR. You should go ahead and merge this and we can address my previous comments in another PR

PR #803, opened 2026-07-17
Add BTLx `UserReferencePlane` support, remove `CutPlaneSpec`/`MiterPlaneSpec`

### LINKED ISSUE #721: add UserReferencePlane functionality in TimberElement
This is a bit T2 related, the idea is to be able to be defining `BTLxProcessings` to custom reference planes instead of the usual reference sides of the `BTLxPart`/`TimberElement`.

This will be used mainly by processings such as `FreeContour`

The idea is to introduce:
- a mechanism where a user can add custom planes and store them in the `TimberElement` attributes with an index-based ID (required by BTLx) which could then be accessed/used/referenced in the `BTLxProcessings`
- a classmethod in `FreeContour` where a `FreeContour` object can be created with a custom user reference plane, that gets added to the `TimberElement` and gets referenced in the processing

papachap:
This PR adds support for BTLx's `UserReferencePlane` concept and uses it to replace the ad-hoc plane-override system joints were using before.

- `TimberElement`s can now register arbitrary named planes via `add_user_ref_plane()` / `get_user_ref_plane()`, stored on `TimberElement.attributes` (that I believe is much underused) relative to the element's `ref_frame`. These round-trip through BTLx XML, survive `transform()`, and survive model (de)serialization.
- `ButtJoint`/`LButtJoint`/`TButtJoint` and `LMiterJoint` now reference an override plane by a plain `int` ID (`butt_plane_id`, `back_plane_id`, `miter_plane_id`) instead of building a `CutPlaneSpec`/`MiterPlaneSpec` object.
- `CutPlaneSpec` and `MiterPlaneSpec` are removed entirely — they each encoded basically the same thing (a plane relative to a beam) in a slightly different way per joint family. One mechanism now covers all of them.
- Reworked a bit ButtJoint's pocket logic, to add a Pocket instead of a Lap in case the butt_plane is overridden since the override plane may not be perpendicular to the cross beam's face, which a Lap can't represent.

### Checklist


chenkasirer (commented):
Generally fine by me but wrote some more high-level take on the recent movements in this system [here](https://github.com/gramaziokohler/compas_timber/issues/804)  Hopefully it'll help us evaluate the alternatives in a more engineer-y fashion.

In the meantime, I would jut push back about using the attributes dict - I think it's too unstructured for this purpose, and should be reserved for user or project specific data, not part of core functionality.

src/compas_timber/base.py
```diff
@@ -10,6 +12,47 @@
...
+    frame : :class:`compas.geometry.Frame`
+        The plane relative to the owning element's ``ref_frame``.
+    ID : int
+        The BTLx integer ID of this plane.
+
+    """
+
+    def __init__(self, frame: Frame, ID: int):
+        super(UserReferencePlane, self).__init__()
+        if type(ID) is not int:
```
chenkasirer: use `id_` instead to not shadow the built-in `id`. all caps is reserved for class attributes

src/compas_timber/base.py
```diff
@@ -383,3 +426,115 @@ def get_dimensions_relative_to_side(self, ref_side_index):
...
+        system BTLx processing parameters are defined in when a custom reference plane is declared.
+
+        Use :meth:`add_user_ref_plane` to add a plane in model (world) coordinates, and
+        :meth:`get_user_ref_plane` to resolve a plane back to model (world) coordinates.
+
+        Returns
+        -------
+        list[:class:`UserReferencePlane`]
+        """
+        return self.attributes.get("user_ref_planes", [])
```
chenkasirer: we should use a separate, explicit, store for user ref planes. this is something we can and should do as library maintainers. the attribute dict is a great, flexible, way to add external data to the instances but doesn't provide any benefit in this case.

src/compas_timber/base.py
```diff
@@ -383,3 +426,115 @@ def get_dimensions_relative_to_side(self, ref_side_index):
...
+        if ID is not None:
+            if any(ID == p.ID for p in self.attributes.get("user_ref_planes", [])):
+                raise ValueError("A reference plane with ID {} already exists. Call remove_user_ref_plane first.".format(ID))
+        else:
+            ID = len(self.attributes.get("user_ref_planes", [])) + 100
+
+        self.attributes.setdefault("user_ref_planes", []).append(UserReferencePlane(frame=local_frame, ID=ID))
+        return ID
+
+    def add_user_ref_plane(self, frame: Frame, ID: int = None) -> int:
```
chenkasirer: we should probably guard here against planes which are completely unrelated to the beam. like a plane whose origin is on the other side of the scene

chenkasirer (approved):
LGTM!
left some comments + documented the different alternatives we went through in https://github.com/gramaziokohler/compas_timber/issues/804

src/compas_timber/base.py
papachap: Hmm yeah it makes sense.. But I am wondering if  `add_user_ref_plane` is the best spot for this. It's generic storage with no idea what the plane will be used for, so there's no principled distance threshold to check against.

The BTLx processing itself will throw a `ValueError` if the geometry can't be used, or a joint using that plane would have its own lazy checks against it. Maybe that's enough already?

Happy to hear your ideas on it.

papachap:
@obucklin I have implemented some changes based on @chenkasirer comments.. please have another look :)

obucklin (changes requested):
Hey @papachap,  I'm glad that you finally implemented this in BTLx. However there is a critical issue when applying this to the Joints with custom plane inputs: 

In order to use this as a design input, you will have to find a way to deal with joint extensions. in hindsight, the old `*_plane_spec`  didn't take into account the extensions, and fails when extensions are added on the start end of the beam. (see the image below) 

Image

This is of course because we use the `ref_sides` for parameterizing the plane and not the `beam.frame` or some other "hard-coded" property. I would fix this by using the beam.frame as the initial reference frame, not the ref_side. (Actually I would still just serialize the `Plane` in the beam local space.)  This may be a case where we need to differentiate between the design-side and fabrication-side of `model.process_joinery` and say that the input is different than the output, in which case the `BTLx.UserReferencePlane` is only a result of joinery used for fabrication.

papachap:
@obucklin Thanks for pointing the bug out! please have another look

src/compas_timber/connections/butt_joint.py
```diff
@@ -200,27 +124,44 @@ def main_beam_ref_side_index(self):
...
+        Otherwise defaults to the cross beam's side closest to the main beam.
+        In case `mill_depth` is set, the plane is offset accordingly to accommodate a pocket/lap of the specified depth.
         """
-        if self._butt_plane_spec is not None:
-            return self._butt_plane_spec.to_plane(self.cross_beam)
+        if self.butt_plane_id is not None:
+            user_plane = self._resolve_user_ref_plane(self.cross_beam, self.butt_plane_id, "butt_plane")
+            if self.mill_depth:
+                main_centerline = point_centerline_towards_joint(self.main_beam, self.cross_beam)
+                user_plane.translate(-main_centerline.direction * self.mill_depth)  # offset the plane along the beam's centerline direction
```
obucklin, line 159: I'm glad you are implementing this! I would think the mill depth would offset normal to the user_plane.

obucklin (changes requested):
One comment in the code. I added thoughts to #804 @chenkasirer s conversation about this topic. I am glad you fixed the extension bug but I see two issues:

1.  Joint doesn't own a joint attribute, meaning e.g. the `LMiterJoint.miter_plane` can be changed in the `TimberElement.add_user_ref_plane()` method. I think the Joints should own those attributes. 
2. you now essentially have 2 `user_plane` types. one is the `UserReferencePlaneCollection` on `TimberElement`, and one is the BTLx user_plane, which is generated via 2 transformations and is a different object. IMO this deviates from the "feature is a `BTLxprocessing`" philosophy of CT.

Since this is a recurring issue, I suggest we sit together and discuss to come to a consensus on how to solve this.

papachap:
@obucklin Thanks for your feedback! I would like to reflect to the two points that you raised:

**1. Joint owning the plane:**
- You have a point there. If for whatever reason a referenced `user_reference_plane` gets removed, then a joint that was making use of it would definitely fail. 
- Probably a locking mechanism for `user_reference_planes` that have been accessed and are in use, would be one way to guard these planes. 

**2. BTLx alignment issue:**
  - This might be a misconception. There aren't two `user_reference_planes`, the API only ever exposes one, and it's always in world coordinates via `.add/.get_user_reference_plane()`. The `UserReferencePlaneCollection` mechanism is what takes care of storing it, by referencing it to the beam's immutable frame rather than its (mutable, blank derived) `ref_frame`; nothing in the public API ever returns the collection class itself.
  - What you're reading as "a dual frame" is just the BTLx serialization step, which re-expresses that plane relative to the element's `ref_frame` so it matches the BTLx schema. But that's no different from the duality between a `TimberElement` and a `BTLxPart`, which also isn't meant to be touched by a user, it's just what the writer uses internally to serialize.  
  - The only genuinely 1-to-1 system we have is `BTLxProcessings`, and even there it's not perfectly clean, since Python and BTLx aren't always fully compatible (see `FreeContour`). 
  One small note: BTLx `UserReferencePlanes` aren't on the same level as `BTLxProcessings` to begin with, they're on the same level as BTLx  `Transformation`/`ReferenceSide`: Part level reference geometry, not a machining operation. So comparing it to the processing 1:1 pattern is comparing it to the wrong thing in the schema.

obucklin:
@papachap fair points. I still think the joint should own their own properties. For simplicity sake I would do this as a transformation from the element_a.frame to the plane. I think we should sit with @chenkasirer and discuss to decide.

PR #810, opened 2026-07-24
Fixes Deseiralization/Serialization bug for children elements.

### LINKED ISSUE #809: Bug report: TimberElement.__data__ serializes the world frame instead of the local/hierarchical frame, causing double-transformation on JSON round-t
### Summary

Any `Beam` or `Plate` (or other `TimberElement` subclass) that is **nested inside a parent with a non-identity transformation** (e.g. a `Panel`, or a `Layer` inside a `Panel`) ends up in the wrong world position after a `json_dumps` → `json_loads` round-trip. The position error is exactly the parent chain's transformation applied a second time.

### Root cause

`TimberElement.__data__` in `src/compas_timber/base.py` (around line 66-75):

```python
@property
def __data__(self):
    data = {}
    data["frame"] = self.frame
    data["length"] = self.length
    data["width"] = self.width
    data["height"] = self.height
    data["features"] = [f for f in self.features if not f.is_joinery]
    data.update(self.attributes)
    return data
```

`self.frame` is inherited from `compas_model.elements.element.Element`:

```python
@property
def frame(self) -> Frame:
    return Frame.from_transformation(self.modeltransformation)
```

`modeltransformation` is the **world** transformation — it's computed by walking the full parent chain (`compute_modeltransformation` in `compas_model`) and multiplying every ancestor's `transformation` together with this element's own.

But `Beam.__init__` / `Plate.__init__` (via `TimberElement.__init__`) treat the `frame` constructor argument as **local**, relative to wherever the element ends up in the hierarchy:

```python
def __init__(self, frame, length, width, height, features=None, **kwargs):
    super().__init__(transformation=Transformation.from_frame(frame), features=features)
```

i.e. `self.transformation = Transformation.from_frame(frame)` — the incoming `frame` is assigned straight to the element's *local* transformation, not the world one. This is also stated explicitly in `Beam`'s docstring: *"frame: The frame representing the beam's local coordinate system in its hierarchical context."*

So there's a mismatch: `__data__` serializes the **global** frame, but `__init__` expects the **local** frame. When an element has no parent, `self.frame == self.transformation` (parent chain is trivial), so the bug is invisible for top-level elements. But for any element nested under a parent with its own transformation, after deserialization:

1. `Beam.__init__` sets `self.transformation` = the **already-global** frame from the old model.
2. The element is re-attached to its (correctly-reconstructed) parent in the tree.
3. `compute_modeltransformation()` walks the parent chain again and multiplies the parent's transformation onto `self.transformation` — which is already global.

Net effect: the parent chain's transformation gets applied twice.

Note: `Panel.__data__` and `Layer.__data__` don't have this problem — they override `__data__` to serialize `self.plate_geometry` directly (which is stored in local coordinates), not `self.frame`. The bug is specific to `TimberElement.__data__` (used by `Beam`, `Plate`, and any other subclass that doesn't override `__data__`'s `frame` field).

### Reproduction

```python
from compas.data import json_dumps, json_loads
from compas.geometry import Frame, Point, Vector
from compas_timber.elements import Beam, Panel
from compas_timber.model import TimberModel

model = TimberModel()
panel = Panel(frame=Frame(Point(10, 0, 0), Vector(1, 0, 0), Vector(0, 1, 0)), length=5, width=3, thickness=0.2)
model.add_element(panel)

beam = Beam(frame=Frame(Point(0, 0, 0), Vector(1, 0, 0), Vector(0, 1, 0)), length=2, width=0.1, height=0.1)
model.add_element(beam, parent=panel)

print("before:", beam.centerline)   # Line(Point(10, 0, 0), Point(12, 0, 0))  -- correct

model2 = json_loads(json_dumps(model))
beam2 = [e for e in model2.elements() if isinstance(e, Beam)][0]
print("after: ", beam2.centerline)  # Line(Point(20, 0, 0), Point(22, 0, 0))  -- WRONG, doubled
```

Expected: `after` matches `before`. Actual: the panel's `x=10` offset gets applied twice, landing at `x=20`.

The same happens (compounded further) with deeper nesting, e.g. `Panel → Layer → Beam`.

### Proposed fix

In `src/compas_timber/base.py`, serialize the local transformation instead of the world frame:

```python
data["frame"] = Frame.from_transformation(self.transformation)  # was: self.frame
```

`Frame` is already imported in that file. This makes `__data__` consistent with what `__init__` expects, so round-tripping a nested `Beam`/`Plate` reproduces its original world position.

ericgozzi:
Fixes #809 

This pull request addresses a critical bug in the serialization logic for `TimberElement` and adds a regression test to ensure nested elements (like a `Beam` inside a `Panel`) preserve their world position after a round-trip through JSON. It also updates the changelog to document the fix.

Bug fix for serialization:

* Updated the `TimberElement.__data__` property in `src/compas_timber/base.py` to serialize the local transformation (using `Frame.from_transformation(self.transformation)`) instead of the world frame, preventing parent transformations from being double-applied during deserialization.

Testing improvements:

* Added a regression test `test_beam_nested_in_panel_roundtrip_preserves_world_position` in `tests/compas_timber/test_panel.py` to verify that a nested `Beam` maintains its correct world position after serialization and deserialization.

Documentation:

* Updated `CHANGELOG.md` to describe the fixed bug with `TimberElement.__data__` and its impact on nested element serialization.

### Checklist

### DIFF
--- a/tests/compas_timber/test_panel.py
+++ b/tests/compas_timber/test_panel.py
@@ -110,6 +110,26 @@ def test_add_beam_to_panel(model):
     assert beam.modeltransformation == model.panels[1].transformation, "Expected beam model transformation to match panel transformation"
 
 
+def test_beam_nested_in_panel_roundtrip_preserves_world_position():
+    # regression test: TimberElement.__data__ used to serialize the world frame instead of the
+    # local transformation, causing the parent panel's transformation to be applied twice on reload.
+    new_model = TimberModel()
+    panel = Panel(frame=Frame(Point(10, 0, 0), Vector(1, 0, 0), Vector(0, 1, 0)), length=5, width=3, thickness=0.2)
+    new_model.add_element(panel)
+
+    beam = Beam(Frame(Point(0, 0, 0), Vector(1, 0, 0), Vector(0, 1, 0)), length=2, width=0.1, height=0.1)
+    new_model.add_element(beam, parent=panel)
+
+    expected_centerline = beam.centerline
+
+    model_copy = json_loads(json_dumps(new_model))
+    panel_copy = next(e for e in model_copy.elements() if isinstance(e, Panel))
+    beam_copy = next(iter(panel_copy.children))
+
+    assert TOL.is_allclose(expected_centerline.start, beam_copy.centerline.start)
+    assert TOL.is_allclose(expected_centerline.end, beam_copy.centerline.end)
+
+
 def test_copy_panel_model(model):
     model_copy = json_loads(json_dumps(model))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed `BallNodeJoint`, `YButtJoint`, and `TOliGinaJoint` not recording all of the features they apply in `self.features`, which meant `clear_features()` (or the old per-joint clearing logic) could leave some features permanently stuck on the beams.
 * Fixed `PlateJoint.clear_extensions()` resetting *all* of an element's extensions when the joint never set one (e.g. `PlateTButtJoint`'s cross plate), instead of leaving unrelated joints' extensions untouched.
 * Fixed panel `Opening` geometry calculations in standalone environments by swapping `compas.geometry.Brep` for `compas_brep`.
+* Fixed `TimberElement.__data__` (used by `Beam`, `Plate`, and other subclasses that don't override it) serializing the world frame instead of the local/hierarchical transformation, which caused the parent chain's transformation to be applied a second time when a nested element (e.g. a `Beam` inside a `Panel`) was round-tripped through JSON.
 
 ### Removed
 * Removed depricated `features.py` module and related imports.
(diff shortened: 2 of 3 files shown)


tests/compas_timber/test_panel.py
```diff
@@ -110,6 +110,26 @@ def test_add_beam_to_panel(model):
...
 
 
+def test_beam_nested_in_panel_roundtrip_preserves_world_position():
+    # regression test: TimberElement.__data__ used to serialize the world frame instead of the
+    # local transformation, causing the parent panel's transformation to be applied twice on reload.
+    new_model = TimberModel()
+    panel = Panel(frame=Frame(Point(10, 0, 0), Vector(1, 0, 0), Vector(0, 1, 0)), length=5, width=3, thickness=0.2)
+    new_model.add_element(panel)
+
+    beam = Beam(Frame(Point(0, 0, 0), Vector(1, 0, 0), Vector(0, 1, 0)), length=2, width=0.1, height=0.1)
```
obucklin, line 120: maybe add a more meaningful position than worldXY for your frame here.

obucklin (commented):
Good catch! just one comment about the test. otherwise LGTM

PR #811, opened 2026-07-27
decouple JointCandidate from Joint 

### LINKED ISSUE #511: Consider making JointCandidate its own thing

### LINKED ISSUE #639: Evaluate the coupling/dependency between: `Analyzer`/`Cluster`/`Joint`/`JointCandidate`/`JointRules`

### LINKED ISSUE #815: Should `JointCandidate`s survive serialization? Deserialized candidates come back element-less and unremovable
While studying the joint lifecycle I noticed joints and joint candidates are stored asymmetrically, and only joints participate in deserialization re-linking. Verified live (repro below) and by inspection on `main` at `148d51b405`.

- `add_joint(joint)` stores the joint in `model._joints` and writes only a guid string to the edge's `"joints"` attribute; on load, `__from_data__` calls `restore_elements_from_keys(model)` for every entry — references re-linked.
- `add_joint_candidate(candidate)` writes the candidate **object itself** to the edge's `"candidates"` attribute (`model.py:425`); there is no `_candidates` dict, and on load nothing re-links candidates.

Since `JointCandidate` is a `Joint` subclass it serializes fine, but after `compas.json_loads` it reconstructs down the deserialization branch of `Joint.__init__` (`joint.py:97-101`) with `_elements = ()`, never restored.

Repro (two crossing beams, `connect_adjacent_beams()`, then `json_dumps` → `json_loads`):

```
candidates before round-trip: 1   (topology: 3, elements: ['Beam', 'Beam'])
candidates after round-trip:  1   (topology: 3, elements: (), interactions: [])
joint after round-trip: TButtJoint elements: ['Beam', 'Beam']   # real joints re-link fine
```

Two consequences of the shell state:

1. Any consumer reading candidates off a reloaded model sees candidates with no elements.
2. The shell can't be removed: `remove_joint_candidate` locates the edge via `candidate.interactions` (`model.py:535`), which is empty — so the clearing pass at the start of `connect_adjacent_beams()` (`model.py:644-646`) is a no-op on a reloaded model, and stale candidates are only ever overwritten or orphaned.

**Question:** are `JointCandidate`s meant to survive serialization at all?

- If **no** (transient detection state, recomputed anyway) — would you welcome a PR that keeps them out of the serialized graph, so a reloaded model starts candidate-clean?
- If **yes** — should `__from_data__` grow a second walk over edge `"candidates"` calling `restore_elements_from_keys`, mirroring the joints loop?

The current middle state (persisted but never restored, removable only before saving) looks like an accident of the storage asymmetry rather than a decision — but I may be missing the intent.

obucklin:
`JointCandidate`/`PlateJointCandidate` no longer subclass `Joint`/`PlateJoint`

## Summary

`JointCandidate` was previously implemented as a `Joint` subclass (and `PlateJointCandidate` as a
`PlateJoint` subclass), even though candidates are information-only: they carry no features, no
fabrication behavior, and can't be processed until *promoted* to a real joint. Inheriting from `Joint`
meant candidates dragged along a large API surface they never used, while also being misleadingly
`isinstance(x, Joint)`.

This PR makes both classes standalone `compas.data.Data` subclasses, implementing only what a
candidate actually needs:
- `elements` / `element_a` / `element_b` / `interactions`
- `topology`
- a lazily-computed `location` (falls back to `location_from_centerlines()`, same as before)
- serialization (`__data__` / `restore_elements_from_keys`), mirroring `Joint`'s existing pattern

`TimberModel.get_candidate(element_a, element_b)` was added as the candidate-side counterpart to the
existing `get_joint()`. Candidates continue to be stored only on the graph edge (under the
`"candidates"` attribute), not in a separate registry — a candidate always connects exactly two
elements, so no extra registry like `TimberModel._joints` is necessary.

Along the way, fixed a real bug (see below) and added test coverage for two previously-untested code
paths.

## Breaking change

Any code doing `isinstance(candidate, Joint)` (or `isinstance(plate_candidate, PlateJoint)`) needs to
switch to `isinstance(candidate, JointCandidate)` / `isinstance(candidate, PlateJointCandidate)`. No
such usage exists inside `compas_timber` itself; `timber_design` should be checked before merging,
per the cross-repo dependency rule (this repo must never import from `timber_design`, so that check
has to happen on the `timber_design` side).

`model.joint_candidates`, `model.unpromoted_joint_candidates`, `model.add_joint_candidate()`,
`model.remove_joint_candidate()`, and graph-edge storage are all unchanged from `main`.

### Checklist


chenkasirer (approved):
LGTM

PR #788, opened 2026-07-10
Make mermaid class diagrams zoomable in docs

daniel-locatelli:
## What

Adds [mkdocs-panzoom-plugin](https://github.com/PLAYG0N/mkdocs-panzoom) so the mermaid class diagrams in the Developer Guide can be zoomed (Alt + mouse wheel), panned (drag), and opened fullscreen. Zoom state persists per diagram.

## Why

The class diagrams on the [Class Diagrams page](https://gramaziokohler.github.io/compas_timber/latest/contribution/class_diagrams/) are large and currently rendered as static SVGs, which makes them hard to read.

## Changes

- `pyproject.toml`: add `mkdocs-panzoom-plugin` to the `dev` extra
- `mkdocs.yml`: enable the `panzoom` plugin with `full_screen: true`
- `uv.lock`: refreshed — also bumps `compas-brep` to 0.1.4, matching the existing `>=0.1.3` requirement (same resolution as #786; whichever merges first, the other becomes a no-op)

No changelog entry per convention for project-file/docs changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #812, opened 2026-07-27
Regenerate class diagrams from source and fix contribution guides

daniel-locatelli:
## Summary

Fixes the four pre-existing class diagrams (Elements, Connections, Fabrication, Errors) so they match the actual v2.2.0 code, and makes them reproducible:

- The diagrams are generated with a pipeline maintained outside this repo at https://github.com/daniel-locatelli/compas_timber-class-diagrams: class bodies, inheritance and attributes are extracted from `src/` with Python's `ast` (ground truth), while subsystem partitioning, prose and cross-class edges are curated in `diagram_spec.py`. Curated composition edges are asserted against the extracted attributes so they fail loudly when the code changes; see that repo's README for usage.
- Regenerates `docs/contribution/class_diagrams.md` from that pipeline; the page header now states it is generated and must not be edited by hand.
- Drops the `Data` base-class nodes (they carried no discriminating information — the prose states the convention and its exceptions instead).
- Draws the constants classes (`OrientationType`, `StepShapeType`, `TenonShapeType`, `AlignmentType`, `EdgePositionType`, `LimitationTopType`) and `MachiningLimits` as explicit blocks with source-verified usage edges — previously the constants classes appeared in the diagrams only as attribute types, and `MachiningLimits` not at all. `BTLxFromGeometryDefinition` gains an `instantiates` edge to `BTLxProcessing` in place of the dropped `Data` link.
- Per review: `CutPlaneSpec` / `MiterPlaneSpec` are left out of the Connections diagram while that mechanism is reworked in #803, and the Fabrication intro links the [BTLx standard](https://design2machine.com/btlx/index.html) specification.
- Fixes broken mermaid rendering on the published site via `mkdocs.yml`: the superfence formatter is switched from mermaid2's non-escaping `fence_mermaid_custom` to `pymdownx.superfences.fence_code_format`, because the non-escaping variant let the browser parse class-diagram stereotypes like `<<abstract>>` as HTML tags and render them as empty `<>`. (The panzoom enablement originally bundled here landed separately via #788; this branch has been rebased onto main accordingly.)
- Rewrites the joints contribution guide to match the v2.2.0 `Joint` base-class API: corrects the `JointTopology` values (the old page misdescribed `TOPO-X`/`TOPO-I` and was missing `TOPO_Y`, `TOPO_K` and the plate topologies), replaces the outdated `elements` / `restore_beams_from_keys()` requirements with the automatic GUID restoration and the `_set_unset_attributes()` hook, documents `check_elements_compatibility()` as a classmethod, and makes the `add_features` example idempotent on re-run. This API is unchanged on current main, so the guide stays correct after the merge; the one thing to revisit for the next release is the example's manual remove-features-first step, which post-2.2.0 code supersedes with the new `Joint.clear_features()`.

Docs-only change, no CHANGELOG entry ("no changelog" label).

## Heads-up: post-2.2.0 hand-edits to the diagrams are superseded

`main` received hand-edits to `class_diagrams.md` documenting post-2.2.0 additions (e.g. `CompositeJoint` from the composite-joint work). This PR replaces the page with the generated 2.2.0 snapshot, so those entries are intentionally not included — the diagrams document the released version, and new classes will re-enter the diagrams when they are regenerated for the next release (the compare script in the [pipeline repo](https://github.com/daniel-locatelli/compas_timber-class-diagrams) reports any unplaced classes). Going forward the page should not be edited by hand; the header now says so.

## Note for maintainers: updating the published 2.2.0 docs

The docs workflow only deploys the mkdocs site on version tags, so merging this PR does not update the live site by itself — the fixes would first appear with the next release tag. If we want the published "2.2.0 (latest)" docs to show the corrected diagrams before then, someone with push rights needs to run, from the merged `main`:

```
mike deploy 2.2.0 latest --push
```

This rewrites only the 2.2.0 folder on `gh-pages` (no git tag is created or moved).


docs/contribution/class_diagrams.md
```diff
@@ -2,437 +2,554 @@
...
       Opening ..> OpeningType : uses
+      PanelConnectionInterface ..> InterfaceRole : uses
 ```
 
 ## Connections Subsystem
 
-The connections subsystem defines joints and their relationships. All joints inherit from the base `Joint` class and are categorized by topology.
+The connections subsystem defines joints and their relationships. All joints inherit from the abstract `Joint` class (a `compas.data.Data` subclass; `Data` is omitted from the diagram since every class here derives from it) and declare the topology they support via `SUPPORTED_TOPOLOGY`. Joints are registered in the `TimberModel` and referenced from the edges of its interaction graph.
+
+Beam joints join two or more `Beam` elements; the generic bases `ButtJoint`, `LapJoint` and `MortiseTenonJoint` share logic across the topology-specific implementations. `CutPlaneSpec` and `MiterPlaneSpec` describe cutting planes relative to a beam's reference side (so they survive model transformations) and are passed to the butt/miter joint constructors via the `butt_plane_spec` / `back_plane_spec` / `miter_plane` parameters.
```
chenkasirer: these are contested and part of a heated discussion [here](https://github.com/gramaziokohler/compas_timber/pull/803). maybe best to leave it out for now

docs/contribution/class_diagrams.md
```diff
@@ -449,146 +566,243 @@ classDiagram
...
+
+      %% Usage relationships
+      ButtJoint ..> CutPlaneSpec : uses
+      LMiterJoint ..> MiterPlaneSpec : uses
 ```
 
 ## Fabrication Subsystem
 
-The fabrication subsystem handles manufacturing features and BTLx processing. All fabrication features inherit from `BTLxProcessing`.
+The fabrication subsystem handles manufacturing features and BTLx processing. All fabrication features inherit from `BTLxProcessing` (a `compas.data.Data` subclass; `Data` is omitted from the diagram); each processing class represents one BTLx machining operation and is instantiated through alternative constructors (e.g. `from_plane_and_beam()`) rather than directly. Several processings also export a lightweight `*Proxy` companion (`JackRafterCutProxy`, `DoubleCutProxy`, `DrillingProxy`, `LapProxy`, `PocketProxy`, `LongitudinalCutProxy`) that defers the expensive parameter computation until the processing is actually applied; proxies mirror the alternative constructors of their processing and are omitted from the diagram.
```
chenkasirer: maybe add a mention here that this is a (partial) implementation of the BTLx standard in Python and link to the specification, if it's not already linked somewhere there.

https://design2machine.com/btlx/index.html

scripts/docs_class_diagrams/README.md
chenkasirer, line 1: mmh dunno if this suite of scripts should end up as part of compas timber. if it can be generalized, perhaps it's worth packaging this as some kind of a developer tool. otherwise, I would just remove it from here.

daniel-locatelli: Agree, I removed the scripts and have them [here](https://github.com/daniel-locatelli/compas_timber-class-diagrams). I've been thinking about how to keep the diagrams from drifting long-term. Ideally they would be generated as part of the docs CI. If you think that makes sense I can write it as an issue.

docs/contribution/class_diagrams.md
daniel-locatelli: Link added

chenkasirer (approved):
LGTM

Issue #829, opened 2026-07-29
Allow a preferred ref side for face-independent processings in the joint-driven flow


daniel-locatelli:
Constructors like `JackRafterCut.from_plane_and_beam` accept `ref_side_index`, but joints construct processings internally — so in the normal workflow there's no way to choose which reference side the BTLx parameters are expressed on. This matters when downstream tooling reads the file off a specific face (e.g. post-processing referenced on the top face); the current workaround is flipping beams / overriding processings manually.

For plane-defined processings (`JackRafterCut`, `DoubleCut`, through-`Drilling`) any ref side describes the same cut, so re-expressing them on a preferred side is always legal. Face-bound ones (e.g. `Mortise` — the side determines which face the pocket is milled into) must keep the joint's choice.

Would you consider a mechanism to set a preferred ref side after features are generated — e.g. a hint consulted at construction/export, or a post-`process_joinery` re-expression pass? Either shape seems to need processings to declare whether they're face-independent, which the API currently doesn't.

Follow-up from the discussion around #824.

obucklin:
Consider implementing as a post-processing step

PR #830, opened 2026-07-29
Resolve TODO in JackRafterCut.from_plane_and_beam: any ref side yields an equivalent cut

### LINKED ISSUE #824: `JackRafterCut.from_plane_and_beam`: is the `ref_side_index=0` (RS1) default arbitrary? (resolving the TODO)
Tiny question about a TODO, asked so the answer can become a sentence in the code. (Checked against `main` at `08411cd788`.)

`JackRafterCut.from_plane_and_beam` (`fabrication/jack_cut.py:170`):

```python
ref_side = beam.ref_sides[ref_side_index]  # TODO: is this arbitrary?
```

`ref_side_index` is a documented keyword parameter, "Default is 0 (i.e. RS1)" (`jack_cut.py:157-158`), and the TODO sits on the line that uses it — so read literally it questions whether defaulting to RS1 is arbitrary. Given a cutting plane, a valid cut can usually be computed from several reference sides, each producing a different `(orientation, start_x, angle, inclination)` tuple; RS1 is a convention rather than a necessity, but the exported parameters depend on it whenever callers don't choose a side (including when processings are resolved from proxies at export time).

**Question:** is the RS1 default a BTLx-spec convention, a machine-friendliness choice, or genuinely arbitrary? Whichever it is, could the TODO become a one-line comment saying so? Happy to PR the comment with your answer.

daniel-locatelli:
Resolves the `TODO: is this arbitrary?` on the `ref_side_index` default (discussed in #824). The answer has two halves: the default *is* an unprincipled pick — but a harmless one, because `JackRafterCut` is plane-defined: parameters computed on any reference side describe the same cut, and the index only pins which side they're expressed on. That fact now lives in the `ref_side_index` docstring (constructor and proxy) instead of the comment. Doc-only change (plus a changelog line).

The discussion also surfaced the real gap: in the joint-driven flow there's no way to pass a *preferred* ref side (needed e.g. when post-processing reads the BTLx off a chosen face). That follow-up is now tracked in #829.

Fixes #824

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added new `compas_timber.fabrication.BirdsMouth`.
 
 ### Changed
+* Documented in `JackRafterCut.from_plane_and_beam` (and its proxy) that the cut is fully defined by the input plane, so `ref_side_index` only pins which reference side the parameters are expressed on; removed the resolved `TODO` (#824).
 * `FeatureApplicationError` raised from `BTLxProcessing.apply()` now carries geometry in the model's global coordinate system (previously local/element space), matching errors raised elsewhere. 
 * Fixed a live crash (`TypeError`) and two other constructor-argument bugs on `BeamJoiningError` call sites.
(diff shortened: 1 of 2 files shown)


obucklin:
@daniel-locatelli feel free to pin me to review PRs as well

obucklin (approved):
LGTM

PR #831, opened 2026-07-29
Add missing docs favicon

daniel-locatelli:
The mkdocs theme config has pointed the favicon at `public/favicon.ico` since the mkdocs setup was introduced, but the file itself was never committed — so the docs site showed no tab icon.

This adds a multi-size favicon (16/32/48/64 px) generated from `docs/_logo/PNG_tranparent-background.png`: cropped to the solid logo bounds, padded with a small symmetric margin, and downscaled with premultiplied-alpha area averaging so the edges stay clean and symmetric at tab size.

No changelog entry — docs-only asset.


chenkasirer (approved):
LGTM

PR #832, opened 2026-07-29
Fix README: logo alignment, DOI badge, broken links, uv install, authors

daniel-locatelli:

- [x] Documentation update

Fixes the README (logo/title alignment, broken DOI badge, dead tutorial and API links, `uv` install instructions incl. the `viz` extra) and makes `CITATION.cff` the single place where authors are listed.

A contributor only adds their name to `CITATION.cff`. Everything downstream is generated:

```mermaid
%%{init: {"flowchart": {"wrappingWidth": 260}}}%%
flowchart TB
    CFF["<b>CITATION.cff</b><br/>hand-edited · the only author list"]:::focal

    CFF --> GH["<b>GitHub</b><br/><i>Cite this repository</i>"]:::out
    CFF --> ZEN["<b>Zenodo</b><br/>read when a release is archived"]:::out
    CFF --> PYPI["<b>PyPI</b><br/><i>release workflow</i>: <span style="font-family:monospace">invoke pre‑build</span> regenerates pyproject.toml authors"]:::out
    CFF --> CITE["<b>Docs citing page</b><br/><i>docs workflow</i>: mkdocs hook fills the BibTeX author block at build time"]:::out

    classDef focal fill:#fde7db,stroke:#eb6c36,stroke-width:1.5px,color:#2d3142
    classDef out fill:#e9ebf0,stroke:#7a8399,stroke-width:1px,color:#2d3142
```

Nothing is committed by CI. The committed `pyproject.toml` may lag `CITATION.cff` between releases; the release build regenerates it before packaging.

🤖 Co-created with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -1,0 +1,20 @@
+# Authors
+
+`compas_timber` is developed by Gramazio Kohler Research and contributors:
+
+* Aleksandra Anna Apolinarska <<apolinarska@arch.ethz.ch>>
+* Chen Kasirer <<kasirer@arch.ethz.ch>>
+* Gonzalo Casas <<casas@arch.ethz.ch>>
+* Jonas Haldemann <<haldemann@arch.ethz.ch>>
+* Oliver Appling Bucklin <<bucklin@arch.ethz.ch>>
+* Aurèle L. Gheyselinck <<gheyselinck@arch.ethz.ch>>
+* Panayiotis Papacharalambous <<papacharalambous@arch.ethz.ch>>
+* Anastasiia Stryzhevska <<astryzhevska@arch.ethz.ch>>
+* Jelle Feringa <<jelleferinga@gmail.com>>
+* Joseph Kenny <<jk6372@princeton.edu>>
+* Beverly Lytle <<lytle@arch.ethz.ch>>
+* Eric Gozzi <<eric.gozzi@arch.ethz.ch>>
+* Rodrigo Arca Zimmermann <<rodrigo.arca@gmail.com>>
+* Nicolas Benjamin Boscoboinik <<boscoboinik@arch.ethz.ch>>
+* Paul Jaeggi <<jaeggipa@ethz.ch>>
+* Daniel Nunes Locatelli <<daniel.nlocatelli@gmail.com>>
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ uv run pytest
 
 * [Documentation](https://gramaziokohler.github.io/compas_timber/)
 * [COMPAS TIMBER Grasshopper Plugin (timber_design)](https://github.com/gramaziokohler/timber_design)
-* [COMPAS TIMBER Grasshopper Tutorial](https://gramaziokohler.github.io/compas_timber/latest/tutorials/)
+* [COMPAS TIMBER Grasshopper Tutorial](https://gramaziokohler.github.io/timber_design/)
 * [COMPAS TIMBER API Reference](https://gramaziokohler.github.io/compas_timber/latest/api/compas_timber.model/)
 
 ## Questions and feedback
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ pip install compas_timber
 ## First Steps
 
 * [Documentation](https://gramaziokohler.github.io/compas_timber/)
+* [COMPAS TIMBER Grasshopper Plugin (timber_design)](https://github.com/gramaziokohler/timber_design)
 * [COMPAS TIMBER Grasshopper Tutorial](https://gramaziokohler.github.io/compas_timber/latest/tutorials.html)
 * [COMPAS TIMBER API Reference](https://gramaziokohler.github.io/compas_timber/latest/api.html)
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 <h1>
-  <img src="docs\_logo\PNG_tranparent-background.png" alt="Logo" width="70" height="70" style="vertical-align:middle"> 
-      COMPAS Timber
+  <img src="docs/_logo/PNG_tranparent-background.png" alt="COMPAS Timber logo" width="70" height="70" align="absmiddle"> COMPAS Timber
 </h1>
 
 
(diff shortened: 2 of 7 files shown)


AUTHORS.md
chenkasirer, line 1: I think it's best if we don't have to maintain two independent author lists. this file was there in the past and got removed. is there something that requires that there's a separate `AUTHORS.md`?

README.md
```diff
@@ -10,26 +9,38 @@
...
+```
+
+For a development setup, clone the repository and sync the environment:
 
 ```bash
-pip install compas_timber
+git clone https://github.com/gramaziokohler/compas_timber.git
+cd compas_timber
+uv sync --extra dev
+uv run pytest
```
chenkasirer, line 41: there's a `viz` flag as well that installs `compas_brep[occ]` + `compas_viewer` for visualization outside of Rhino. perhaps worth mentioning here as well.

AUTHORS.md
daniel-locatelli: I just checked and the only reference to AUTHORS.md was a link in CONTRIBUTING.md.

While checking I noticed that there are multiple disconnected places storing author lists and they have drifted: citing.md has 11 names vs 16 in pyproject.toml, and Zenodo auto-fills from GitHub usernames.

There is a possible approach to having a canonical [CITATION.cff](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files) at the root of the project. Both GitHub "Cite this repository" button and the Zenodo integration read it.

The issue left would be how to keep CITATION.cff and pyproject.toml in sync. Claude gave two suggestions:

1. Make CITATION.cff the single canonical people list, slim pyproject down. This is what compas core effectively does: its pyproject has one author (Tom Van Mele) while the human-facing list lives elsewhere. compas_timber's pyproject could carry just "Gramazio Kohler Research", and every person-level credit — names, ORCIDs, affiliations — lives in CITATION.cff. One list, one pointer. CFF is the right canonical format anyway since it's the richer schema; you can't round-trip ORCID/affiliation through pyproject.
2. Keep both, add a drift check. A few lines in CI (or the existing invoke lint) asserting every pyproject author appears in CITATION.cff. cff convert already validates the cff file itself. Works, but it's machinery to maintain two lists that option 1 makes unnecessary.

What do you think?

chenkasirer: mmh `CITATION.cff` as a single-source-of-truth sounds like it could be a good solution. maybe that's what we should do.

a good solution for me:
1. consists ideally of just one list of authors that we have to maintain
2. serves the following: Zenodo, Github, mkdocs and PyPi 

So `CITATION.cff` covers significant ground. PyPi still needs pyproject.toml but perhaps we could have an automated task, as part of the publishing pipeline, that scrapes `CITATION.cff` and populates the `pyproject.toml` authors list.

One thing I just notice: PyPi only shows the first author on the list as the package author. this seems to be a known issue documented [here](https://github.com/pypi/warehouse/issues/12877). It seems the solution is to omit the email from the names in the `pyproject.toml` authors list.

So I'd suggest:
* go ahead and create a `CITATION.cff`
* write an invoke task that scrapes it and populates the list of authors in `pyproject.toml` using names only
* use the `pre-build` task to trigger this scraping as part of the publishing pipeline

What do you think?

daniel-locatelli: Sounds good! Working on it

daniel-locatelli: **Done**

- CITATION.cff added
- invoke sync-authors scrapes only the names into pyproject.toml (per the [issue you mentioned](https://github.com/pypi/warehouse/issues/12877)) and regenerates the BibTeX in docs/citing.md. A project-level pre-build task calls it.
- ran it once and committed the result

**Decisions that I need from you**

Our release.yml never actually triggers pre-build. `compas-actions.publish@v3` only runs it on the Grasshopper-components branch, not on the plain PyPI path. Options to guard against future drift:
  -  a small path-filtered check in pr-checks.yml (runs only when CITATION.cff, pyproject.toml or docs/citing.md change, ~15 s)
  - longer-term: teach compas-actions.publish to run pre-build on its plain path

Zenodo's v2.2.0 record includes Matthias Helmreich, who isn't in pyproject.toml. Should he go into CITATION.cff?

README.md
daniel-locatelli: Done 👍

docs/citing.md
```diff
@@ -20,12 +20,17 @@ If you use COMPAS Timber for your work, please acknowledge it by citing it with:
         Stryzhevska, Anastasiia and
         Feringa, Jelle and
         Kenny, Joseph and
-        Lytle, Beverly
+        Lytle, Beverly and
```
chenkasirer, line 23: can this not somehow embed the .cff file? it's yet another list of authors.

daniel-locatelli: it's been generated using `invoke sync-authors`, would it be better to have it as a hook triggered by `mkdocs build` instead? See [this comment](https://github.com/gramaziokohler/compas_timber/pull/832#discussion_r3752564210) in the thread above

chenkasirer: yup, mkdocs hook triggering scraping of citting file on docs generation sounds good to me

AUTHORS.md
chenkasirer: we released a new set of CI actions a couple weeks ago, they're already in `main`. 
merge `main` into your branch and then you should be able to use the new prepare release workflow which can run the prebuild task.
for reference, take a look at how compas_pb triggers generation of the protobuf files on release here: https://github.com/compas-dev/compas_pb/blob/8c33c65bbcd85dcb128e72eda5068483e7ea0a48/.github/workflows/release.yml#L39

what i'd like to achieve is that a new contributor only has to put their name in citation.cff and that's it. everything else happens automatically on release and/or on docs deployment. one shouldn't have to manually run the sync task and commit as part of their PR. If that's not achievable with a reasonable effort / amounts of change to the library, we could add a special CI workflow that just deals with that (e.g. triggers when citation.cff is edited on main).

and yes, Helmreich should be in it.

daniel-locatelli: I rebased to main as we discussed. The sync plumbing necessary for an automatic deployment was already in this PR, it just needed the release.yml. The only edit on top of main is making [sync_authors](https://github.com/gramaziokohler/compas_timber/pull/832/changes#diff-2e746c9eacad6e2c2fdefdc6a665b5ce4607ea384ebf97a729044b79acc778d6R20) a prerequisite of pre_build.

Helmreich added 👍

docs/citing.md
daniel-locatelli: Nice, I added a [mkdocs_hooks.py ](https://github.com/gramaziokohler/compas_timber/pull/832/changes#diff-96e30b390974784d0fdb81cf313f7f276328406652e2709d76a5828ed7cafb44R1)

PR #828, opened 2026-07-29
Link Grasshopper plugin to timber_design repository in docs

daniel-locatelli:
As discussed in #825: the Grasshopper components now live in the separate [`timber_design`](https://github.com/gramaziokohler/timber_design) package, but the docs never pointed there. This adds the link wherever the Grasshopper plugin is mentioned, plus a dedicated page:

- **New page: *Grasshopper Plugin → Development*** (`docs/tutorials/grasshopper/development.md`) — explains that the components are developed in `timber_design` (with a direct link to the components folder), which issue tracker to use for what, that the Package Manager package is still published under the name `compas_timber`, and a mermaid diagram of how `timber_design` fits in the COMPAS ecosystem
- `docs/index.md` — the landing page sentence advertising the Grasshopper plug-in now links to the `timber_design` repository (also fixes an "and and" typo in the same paragraph)
- `README.md` — new entry under *First Steps* linking to the `timber_design` repository
- `docs/tutorials/grasshopper/installation.md` — note at the top of the installation guide pointing to `timber_design`

Docs-only change, so no CHANGELOG entry.

Closes #825.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #825, opened 2026-07-27
Docs and Yak workflow still describe the Grasshopper plugin migrated to `timber_design` - staging or drift?


daniel-locatelli:
Question about docs/CI follow-through on the Grasshopper migration — this may be deliberate staging while timber_design ramps up, so asking before proposing anything. (Checked against `main` at `08411cd788`.)

Since `24a841d050` removed the GH components, `compas_timber.design`, and `compas_timber.ghpython` (migrated to `timber_design`, per the CHANGELOG), a few user-facing surfaces still describe the pre-migration world:

- `docs/index.md` still advertises compas_timber "as a Grasshopper plug-in", and the docs' only tutorials are the 12-page "Grasshopper Plugin" section (`mkdocs.yml:162-175`).
- `docs/tutorials/grasshopper/installation.md` walks users through Rhino 8's PackageManager → install `compas_timber` → "you should see the COMPAS Timber section in the Grasshopper toolbar". Since the Yak publish workflow can't currently run (below), that package is frozen at a pre-migration build, so a new user following the tutorial gets a toolbar that doesn't match the current 2.2.0 docs — and nothing anywhere says the components now live in timber_design (`git grep -i timber_design` → CHANGELOG only).
- `docs/installation.md:36` still instructs contributors to run `invoke build-cpython-ghuser-components`, a task no longer in `tasks.py`.
- `.github/workflows/yak_publish.yml:41-49` references that missing task plus the deleted paths `src\compas_timber\ghpython\components_cpython\ghuser` and `...\yak_template`.

Meanwhile timber_design is public and release-active with its own `yak_publish.yml` — so the successor story clearly exists.

**Questions:**

1. Is the docs silence deliberate staging (waiting for timber_design's docs/Yak story to be user-ready) or just drift?
2. Is the `compas_timber` Yak package meant to be superseded by a timber_design package, or updated to ship the new components under the old discoverable name?
3. Should the GH tutorial section move to timber_design's docs, or stay here pointing across?

If it's drift, happy to PR the small version: a "GH components moved to timber_design" note in `docs/index.md`/README and atop the GH tutorials, delete the stale contributor instructions, and remove or retarget `yak_publish.yml`.

obucklin:
add link to timber_design repo

chenkasirer:
Good idea to add a link. I'd also migrate the plugin documentation (after it's been updated) to the timber design repo and publish a documentation website over there. then add a link to that on Food4Rhino, if not already there.

PR #834, opened 2026-07-30
Fix stale text in Grasshopper tutorial docs, add Under-the-hood tips, fix contribution-guide anchors

daniel-locatelli:
Text-only cleanup of the docs, three concerns:

**Grasshopper tutorial fixes**
- Removes the pre-2.x `compas_timber.parts.Beam` namespace from `beams.md` — GH users don't need Python paths, a beam is just a Beam.
- `show.md`: fixes a typo, replaces leftover "assembly" wording with "model", and un-swaps the `Show Beam Face Index` / `Show Beam Index` descriptions, which had each other's text.

**"Under the hood" tips**
We decided to remove inline Python paths from the Grasshopper tutorials: GH users don't need them, and for that audience they are more confusing than helpful — a `compas_timber.parts.Beam` is just a beam. For readers who do want to dig deeper, the pages with a Python counterpart now end with a one-line tip linking the GH concept to it (`Beam`, `TimberModel`, the fabrication processings, `BTLxWriter`) — including pointers to [timber_design](https://github.com/gramaziokohler/timber_design) where the components actually live (joint rules, design), complementing #828.

**BTLx contribution guide anchors**
The 15 API deep-links pointed at anchors mkdocstrings never generates (e.g. `#DoubleCut.planes_from_params_and_beam`), so they landed at the top of the API page. Now qualified with the full ID format (`#compas_timber.fabrication.DoubleCut.planes_from_params_and_beam`); all 15 verified against the rendered page.

Note: the GH tutorials also reference component names predating the current `timber_design` set (e.g. `Show Beam Index` vs today's `CT_ShowElementIndex`). This is not yet addressed anywhere — #828 only adds signpost links — so it's left out of this PR — whether the tutorials should be updated here or move next to the components is a question worth settling on #825 first.

### DIFF
--- a/docs/contribution/BTLx_contribution_guide.md
+++ b/docs/contribution/BTLx_contribution_guide.md
@@ -72,11 +72,11 @@ class NewProcessing(BTLxProcessing):
 
 See also:
 
-- [`JackRafterCut`](../api/compas_timber.fabrication.md#JackRafterCut)
-- [`Lap`](../api/compas_timber.fabrication.md#Lap)
-- [`Drilling`](../api/compas_timber.fabrication.md#Drilling)
-- [`DoubleCut`](../api/compas_timber.fabrication.md#DoubleCut)
-- [`FrenchRidgeLap`](../api/compas_timber.fabrication.md#FrenchRidgeLap)
+- [`JackRafterCut`](../api/compas_timber.fabrication.md#compas_timber.fabrication.JackRafterCut)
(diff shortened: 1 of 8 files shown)


chenkasirer (approved):
Generally LGTM, but perhaps we should migrate the documentation of the plugin to the [design tool repo](https://github.com/gramaziokohler/timber_design/)
And refreshen it also, I see a few other outdated parts

PR #836, opened 2026-07-30
Delegate `TimberElement.add_feature` to `add_features`

### LINKED ISSUE #823: TimberElement.add_feature silently accepts a list and corrupts the feature list
**Describe the bug**

`add_feature` appends its argument without checking it. Passing a list (which the docstring's "Adds one or more features" wording invites) nests that list inside `_features`. Nothing fails at the call site — the error only appears later, when geometry is computed, far from the mistake.

**To Reproduce**

1. Context: plain Python; same behavior in Rhino 8 Grasshopper (script component with a list-access input).
2. Sample script:

```python
from compas.geometry import Point
from compas_timber.elements import Beam
from compas_timber.fabrication import JackRafterCut

beam = Beam.from_endpoints(Point(0, 0, 0), Point(1, 0, 0), 0.06, 0.12)
beam.add_feature([JackRafterCut(start_x=0.5)])  # no error
print(beam._features)  # [[<JackRafterCut>]] -- nested list
beam.geometry
```

3. See error: `AttributeError: 'list' object has no attribute 'apply'`, raised from `compute_elementgeometry` — the traceback points at the geometry code, not at the bad `add_feature` call. `debug_info` stays empty (the feature loop only catches `FeatureApplicationError`).

**Expected behavior**

Either the list is handled, like `add_features` already does, or the call fails immediately with a `TypeError`.

**Screenshots**

Image

In Grasshopper the component calling `add_feature` stays green; a component reading `.geometry` two wires downstream turns red.

**Desktop (please complete the following information):**

- OS: Windows 11
- Python version: 3.14
- Python package manager: uv (pip); compas_timber 2.2.0, same code on `main` at `fcfccbd4db`

**Additional context**

`add_feature` exists to shadow `compas_model.Element.add_feature` so features land in timber's `_features` storage (#545). Proposed fix: delegate to `self.add_features(feature)` — one storage path, and the list handling comes for free. The stricter alternative (raise on list input) would make the override stricter than the base-class method it shadows. Happy to PR the delegation. The shipped GH components are not affected (they pass single features or use `add_features`); this bites anyone scripting against the API. The docstring fix ("Adds one or more features" on the singular method) would ride the same PR.

daniel-locatelli:
Fixes #823 — `add_feature` appended its argument unchecked, so a list would nest inside `_features` and only fail later, in geometry computation. It now delegates to `add_features`, which normalises single-or-list input. Docstring corrected (it was copied from `add_features`), and the cache-reset decorators dropped since the delegated call carries the same stack. Repro and full context on the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 * Documented in `JackRafterCut.from_plane_and_beam` (and its proxy) that the cut is fully defined by the input plane, so `ref_side_index` only pins which reference side the parameters are expressed on; removed the resolved `TODO` (#824).
+* `TimberElement.add_feature` now delegates to `add_features`, so a list passed by mistake is normalised instead of silently nested (#823). Docstring corrected accordingly.
 * `FeatureApplicationError` raised from `BTLxProcessing.apply()` now carries geometry in the model's global coordinate system (previously local/element space), matching errors raised elsewhere. 
 * Fixed a live crash (`TypeError`) and two other constructor-argument bugs on `BeamJoiningError` call sites.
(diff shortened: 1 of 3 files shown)


chenkasirer (approved):
LGTM

PR #839, opened 2026-08-04
Replace deprecated `element_by_guid` call in `Joint.restore_elements_from_keys`

daniel-locatelli:
`Joint.restore_elements_from_keys` still calls the deprecated `element_by_guid`, so deserializing any jointed model emits a `DeprecationWarning` from inside the library. #661 converted all internal callers to `model[guid]`; the base-class implementation landed two days later from a parallel branch and reintroduced the call. This applies the same conversion — pure passthrough, no behaviour change.

Tests: `test_element_by_guid_deprecated_warning` now actually asserts the warning via `pytest.warns` (the spy only checked the call happened), and a new round-trip test under `error::DeprecationWarning` guards against reintroduction — verified red without the fix, green with it.


chenkasirer (approved):
LGTM

PR #354, opened 2024-12-30
BTLx refactoring & Fabrication package cleanup

papachap:
This PR tackles the refactoring we’ve been discussing regarding the `BTLx` writing process and includes a general cleanup of the `fabrication` package. 

Don’t panic with the amount of files changed! Most of the changes in this PR are simple updates to variables, class names, and docstrings, not major rewrites.

To summarize: 

- I’ve unified terminology by refactoring all `process` references to `processing` 

- Refactored and merged the **btlx module**, including renaming the `BTLx` class to `BTLxWriter`. 

-  I’ve moved around, merged and cleaned up the **fabrication package** by removing deprecated modules like `compas_timber._fabrication` and unused components such as `btlx_processes `and `joint_factories`.

### What type of change is this?

- [ ] Bug fix in a **backwards-compatible** manner.
- [x] New feature in a **backwards-compatible** manner.
- [ ] Breaking change: bug fix or new feature that involve incompatible API changes.
- [x] Other (e.g. doc update, configuration, etc)

### Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [x] I ran all tests on my computer and it's all green (i.e. `invoke test`).
- [x] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [x] I added new functions/classes and made them available on a second-level import, e.g. `compas_timber.datastructures.Beam`.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added necessary documentation (if appropriate)

PR #840, opened 2026-08-04
Record compas_timber version in BTLx FileHistory ProgramVersion

daniel-locatelli:
Exported BTLx files name `ProgramName="COMPAS_Timber"` but stamp `ProgramVersion` with the version of compas, the geometry kernel — so the fabrication artifact can't identify the version of the program that generated its geometry. The string dates to the initial fabrication PR (#138) and survived the BTLxWriter rewrite (#354) verbatim; it was never a decision. This stamps both, compas_timber first: `COMPAS Timber: 2.2.0; COMPAS: 2.15.1`.

Also updates the `test_btlx_file_history` assertion (verified red without the fix, green with it) and fixes a `file_history_attibutes` typo in the same block.


chenkasirer (approved):
LGTM

Issue #821, opened 2026-07-27
`find_intersecting_pairs` docstring describes intersection; implementation is proximity/all-pairs


daniel-locatelli:
Low priority — nothing is broken; this is docstring accuracy plus a naming question. (Checked against `main` at `08411cd788`.)

`ConnectionSolver.find_intersecting_pairs` (`connections/solver.py:107`) is one line:

```python
return find_neighboring_elements(beams, inflate_by=max_distance) if rtree else itertools.combinations(beams, 2)
```

It never tests intersection:

- `rtree=True` → AABB-overlap search over bounding boxes (proximity, not intersection).
- `rtree=False` (the signature default) → every pair, geometry never consulted — and `max_distance` is accepted but **silently ignored** in this mode.

The actual intersection test is downstream in `find_topology`, which returns `TOPO_UNKNOWN` for pairs that don't meet. The method's real job is candidate generation for that check; its only in-`src/` caller is `compute_topologies` (`model.py:731`, always `rtree=True`).

Concrete docstring inaccuracies:

1. "Finds pairs of intersecting beams" — false in both modes (neighboring at best, all pairs by default).
2. Return type promises `list(set(Beam, Beam))`; the default path yields an iterator of tuples.
3. `max_distance` documented without noting it is a no-op when `rtree=False`.
4. Typo "neightboring" here and in `utils/r_tree.py`.

Reader-facing symptom: someone budgeting geometry cost assumes intersection work happens here; by default nothing geometric happens at all — 300 beams "succeed" instantly with 44,850 "intersecting" pairs — and the name hides where real filtering lives, which matters when debugging a missing joint.

**Proposal / questions:**

1. I'd be happy to PR a doc-only fix (rewrite the docstring to say what it does, fix the return type, state the `max_distance`/`rtree=False` interaction, fix the typo in both files). Welcome?
2. Is `rtree=False` still a supported mode, or a leftover from before `rtree` was a hard dependency? If the only sanctioned caller always passes `True`, removing the flag might be more honest than documenting it.
3. Longer-term, is a rename toward the library's own candidate vocabulary (e.g. `find_candidate_pairs`, with the old name deprecated) something you'd consider, or is the current name load-bearing history?

daniel-locatelli:
Rename is off the table and the consolidated neighbor-check work supersedes this. Closing.

PR #841, opened 2026-08-04
Fix drifted docstrings: `TimberModel` accessors, joint lifecycle sentence, solver candidate pairs

daniel-locatelli:
Doc-only sweep, no behavior change.

- `TimberModel` attributes block: the element accessors have returned lists since the `find_all_elements_of_type` refactor, but the class docstring still said Generators (the inline `# type:` comments were already updated). Also adds the missing `fasteners` and `unpromoted_joint_candidates` entries.
- The sentence "This method is automatically called when joint is created by the call to `Joint.create()`" predates `process_joinery()` and survived in 20 docstrings across 12 connections files; replaced with "called during `TimberModel.process_joinery()`, not when the joint is created" — phrased so it stays true for `CompositeJoint` sub-joints, which are never added to the model.
- `find_intersecting_pairs`: the docstring half of #821 (the rename stays off the table) — it generates candidate pairs rather than testing intersection, and `max_distance` is silently ignored when `rtree=False`. Also fixes the "neightboring" typo here and in the two `find_neighboring_elements` plugins.
- `BTLxProcessingError` is collected by `BTLxWriter` and exposed via its `errors` property, never raised — the summary line said otherwise (the class's own third TODO knew this, with a `TOOD` typo, now fixed).
- `register_type_serializer` takes a type *name* string, not a type — both in-repo registrations pass `__name__`, and this matches its sibling `register_type_deserializer`.
- `add_features`/`remove_features`: "beam" → "element", follow-up to #836.

### DIFF
--- a/src/compas_timber/base.py
+++ b/src/compas_timber/base.py
@@ -194,7 +194,7 @@ class TimberElement(Element, abc.ABC):
     @reset_timber_attrs
     def add_features(self, features):
         # type: (BTLxProcessing | list[BTLxProcessing]) -> None
-        """Adds one or more features to the beam.
+        """Adds one or more features to the element.
 
         Parameters
         ----------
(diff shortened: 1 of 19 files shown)


chenkasirer (approved):
LGTM

PR #843, opened 2026-08-10
Fix compas-brep API use in Lap and Pocket

ericgozzi:
Pocket, Lap, and BTLxPart's Brep-handling code still called the old compas.geometry.brep API (face.nurbssurface, face.frame_at) which no longer exists on compas_brep's BrepFace. face.surface now returns a Plane directly for planar faces, so no NurbsSurface/frame_at detour is needed; only genuinely curved faces need surface.frame_at.

Along the way, face.surface didn't account for face.is_reversed, so opposite faces of a box reported identical normals instead of opposite ones - fixed by flipping the normal for reversed faces.

Also fixes _get_optimal_ref_side_index in Pocket and Lap, where edge.curve already returns a Line directly (no .points attribute to unpack).

Pocket.apply additionally now raises a clear FeatureApplicationError when start_depth is negative (the pocket volume lies entirely outside the element's material on the ref_side's outward side) instead of letting a corrupted/erased geometry reach the boolean subtraction.

### Checklist


src/compas_timber/fabrication/btlx.py
```diff
@@ -652,7 +652,14 @@ def shape_strings(self):
             scaled_geometry = self.element.geometry.scaled(self._scale_factor)
             for face in scaled_geometry.faces:
                 pts = []
-                frame = face.surface.frame_at(0.5, 0.5)
```
chenkasirer, line 655: Hey! I think these should be fixed upstream. could you check if https://github.com/gramaziokohler/compas_brep/pull/8
solves these issues?

it should be as easy as calling `face.from_at()` which should now get you the properly oriented frame.

pyproject.toml
```diff
@@ -183,6 +192,10 @@ docstring-code-line-length = "dynamic"
 # Testing
 # ============================================================================
 
+# TEMPORARY -- remove before this reaches main.
+[tool.uv.sources]
+compas_brep = { git = "https://github.com/gramaziokohler/compas_brep", branch = "at_frame" }
```
chenkasirer: is now released so this can be removed

chenkasirer (approved):
LGTM! 
if possible, test again please since there's been much movement now in the compas_brep department.
also please don't forget to remove the branch dependency.

Ah and fix the build ;)

PR #852, opened 2026-09-01
Fix/simple scarf compas brep

paulocinco:
Two fixes to `SimpleScarf` (`compas_timber/fabrication/simple_scarf.py`): the Brep work now runs on `compas_brep`, and `StartX` for END cuts is calculated from the blank instead of carrying a geometry offset in the BTLx parameter.

#### 1. compas_brep

`SimpleScarf` now takes its `Brep` from `compas_brep`, which behaves differently from `compas.geometry.Brep` in two ways the previous code relied on:

* **`from_boolean_difference` returns a list**, one Brep per resulting piece, and `[]` when nothing is left — where `compas.geometry.Brep` returned a single Brep and raised `IndexError` on an empty result. The old code fed that list straight back in as the operand for the drill-hole subtractions, which is not a valid Brep argument. The piece containing the beam centerline midpoint is now selected first, and the drill holes are subtracted from that piece.
* **`from_mesh` needs a real `Mesh`.** `compas_brep` builds one Brep face per mesh face and iterates `mesh.faces()` / `face_vertices()`, which `Polyhedron` does not provide, so the scarf volume is handed over as `scarf_volume.to_mesh()`. Side
  benefit: the quad faces of the scarf volume survive as quads instead of being fanned into triangles the way `Brep.CreateFromMesh` did.

#### 2. StartX / length calculation

`_calculate_start_x` returned `beam.length + length/2` for END orientation. The half-scarf overshoot — needed so the cutting volume clears the end of the beam — was baked into the `StartX` **BTLx parameter**, so the exported value did not point
at the end of the blank and the machine would read a scarf that starts half its own length past the workpiece.

* `_calculate_start_x` now returns `beam.blank_length`, which is the BTLx reference for an END cut (measured along the blank, not the beam).
* The overshoot moved into the geometry, where it belongs: `_planes_from_params_and_beam` translates `top_frame` by `self.start_x` for both orientations, and `blank_plane` (`start_frame + length/2`) supplies the overshoot past the beam end.

### Checklist


chenkasirer (commented):
Generally LGTM, but please take a look at https://github.com/gramaziokohler/compas_timber/pull/855.
merging it first will help your PR with the failing tests. consider stacking this PR onto #855

chenkasirer (approved):
LGTM!

PR #855, opened 2026-09-02
Migrate to compas_brep 0.3.0 boolean operations returning lists

chenkasirer:
> This PR make explicit something that would happen implicitly before. the decision to grab the first boolean result should be revised on a per case basis and tested on each of the backends.

Previously, the rhino backend (also in `compas.geometry.Brep`) would arbitrarily return the first result out or the list of resulting breps from a boolean operations. this was done to match the OCC backend returning just one result. However, boolean operation returning several results is a valid outcome and in OCC is handled by wrapping the results with a COMPOUND solid, which is the reason for the confusion and single return value.

Since `compas_brep==0.3.0`, boolean operations always return a list and it is up to the client to choose the right one. If more than one result exists, the order is backend dependent. 

This PR bumps the minimum dependency in compas brep to the new behavior and includes helper functions that make this choice explicit. I opted for no compatibility wrapper since the move to `compas_brep` is still unreleased.

### Checklist


ericgozzi (approved):
LGTM!

